PR #5, opened 2022-08-09
Let `get_output_status` use a `u64` for the index.

tnull:
Unfortunately #4 had a minor bug as the `get_output_status` API method doesn't require a full `struct Vout` as input and doesn't return the data to populate a full `struct Vin`.

This PR fixes this by just using `u64` for the respective index fields.

Now tested to work with both the `blocking` and `async` variant.


vladimirfomene (commented):
Review ACK. This looks good to me

vladimirfomene (approved):

PR #6, opened 2022-08-10
Add `get_tip_hash` API call

tnull:
This PR adds support for getting the `BlockHash` of the current blockchain tip. 

While this is not time-pressed from our side, it might be nice to have in the future.

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -12,11 +12,12 @@
 //! Esplora by way of `reqwest` HTTP client.
 
 use std::collections::HashMap;
+use std::str::FromStr;
 
 use bitcoin::consensus::{deserialize, serialize};
 use bitcoin::hashes::hex::{FromHex, ToHex};
 use bitcoin::hashes::{sha256, Hash};
-use bitcoin::{BlockHeader, Script, Transaction, Txid};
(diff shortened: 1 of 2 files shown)


vladimirfomene:
@tnull , is it not better to have a method for getting the block hash given a particular block height? I think it will provide a more general API. This is what we currently do here: https://github.com/bitcoindevkit/bdk/blob/master/src/blockchain/esplora/reqwest.rs#L121

tnull:
> @tnull , is it not better to have a method for getting the block hash given a particular block height? I think it will provide a more general API. This is what we currently do here: https://github.com/bitcoindevkit/bdk/blob/master/src/blockchain/esplora/reqwest.rs#L121

Well already have that API, but it may have some drawbacks in certain scenarios: as it is currently implemented, it requires to make two requests (one to get current tip height, one to get the hash), and could even be a bit race-y (i.e., when we make the second call, the block height could just have increased so we don't get the tip, but tip-1).

Generally I think `EsploraClient` should strive to implement all of [Esplora's API methods](https://github.com/blockstream/esplora/blob/master/API.md), allowing a user to pick-and-choose what's required. What of these methods then should be used by/exposed through BDK is a different discussion.

vladimirfomene (approved):

PR #8, opened 2022-09-09
Add `get_txid_at_block_index` API method

tnull:
This PR adds the capability to retrieve a `Txid` by block hash and index, i.e., [this API call](https://github.com/blockstream/esplora/blob/master/API.md#get-blockhashtxidindex)

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -79,6 +79,30 @@ impl AsyncClient {
         }
     }
 
+    /// Get a [`Txid`] of a transaction given its index in a block with a given hash.
+    pub async fn get_txid_at_block_index(
+        &self,
+        block_hash: &BlockHash,
+        index: usize,
+    ) -> Result<Option<Txid>, Error> {
+        let resp = self
+            .client
+            .get(&format!(
+                "{}/block/{}/txid/{}",
+                self.url,
+                block_hash.to_string(),
+                index
+            ))
+            .send()
+            .await?;
+
+        if let StatusCode::NOT_FOUND = resp.status() {
+            return Ok(None);
+        }
+
+        Ok(Some(deserialize(&Vec::from_hex(&resp.text().await?)?)?))
+    }
+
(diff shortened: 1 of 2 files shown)


src/async.rs
```diff
@@ -79,6 +79,30 @@ impl AsyncClient {
...
+                index
+            ))
+            .send()
+            .await?;
+
+        if let StatusCode::NOT_FOUND = resp.status() {
+            return Ok(None);
+        }
+
+        Ok(Some(deserialize(&Vec::from_hex(&resp.text().await?)?)?))
```
afilini, line 103: I think you can simplify this by using `Txid::from_str()`: https://docs.rs/bitcoin/0.29.1/bitcoin/hash_types/struct.Txid.html#impl-FromStr-for-Txid

src/blocking.rs
```diff
@@ -84,6 +84,34 @@ impl BlockingClient {
...
+            .get(&format!(
+                "{}/block/{}/txid/{}",
+                self.url,
+                block_hash.to_string(),
+                index
+            ))
+            .call();
+
+        match resp {
+            Ok(resp) => Ok(Some(deserialize(&Vec::from_hex(&resp.into_string()?)?)?)),
```
afilini, line 104: Same thing about using from_str

src/async.rs
tnull: Ah, good point! Fixed with https://github.com/bitcoindevkit/rust-esplora-client/pull/8/commits/24ccf9cacd5d54f933f7c2a10bc2275beae0a22e.

src/blocking.rs
tnull: Fixed with https://github.com/bitcoindevkit/rust-esplora-client/pull/8/commits/24ccf9cacd5d54f933f7c2a10bc2275beae0a22e.

afilini (approved):
ACK 24ccf9cacd5d54f933f7c2a10bc2275beae0a22e

PR #10, opened 2022-09-16
Add feature (parity) tests for API methods

tnull:
This PR adds basic feature (parity) tests for the API methods:

- [x] `get_tx` 
- [x] `get_tx_no_opt`
- [x] `get_tx_at_block_index`
- [x] `get_tx_status`
- [x] `get_header`
- [x] `get_merkle_proof`
- [x] `get_output_status`
- [x] `get_height`
- [x] `get_tip_hash`
- [x] `get_fee_estimates`
- [x] `scripthash_txs`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,10 @@ reqwest = { version = "0.11", optional = true, default-features = false, feature
 
 [dev-dependencies]
 serde_json = "1.0"
+tokio = { version = "1.20.1", features = ["full"] }
+electrsd = { version = "0.19.1", features = ["legacy", "esplora_a33e97e1", "bitcoind_22_0"] }
+electrum-client = "0.10"
+lazy_static = "1.4.0"
 
 [features]
 default = ["blocking", "async", "async-https"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -198,24 +198,29 @@ impl_error!(bitcoin::hashes::hex::Error, Hex, Error);
 #[cfg(test)]
 mod test {
     use super::*;
-    use bitcoin::Amount;
-    use electrsd::{
-        bitcoind, bitcoind::bitcoincore_rpc::bitcoincore_rpc_json::AddressType,
-        bitcoind::bitcoincore_rpc::RpcApi, bitcoind::BitcoinD, ElectrsD,
-    };
-    use electrum_client::ElectrumApi;
+    use electrsd::{bitcoind, bitcoind::BitcoinD, ElectrsD};
     use lazy_static::lazy_static;
     use std::env;
-    use std::sync::{Mutex, Once};
-    use std::time::Duration;
+    use tokio::sync::Mutex;
+    #[cfg(all(feature = "blocking", any(feature = "async", feature = "async-https")))]
+    use {
+        bitcoin::Amount,
+        electrsd::{
+            bitcoind::bitcoincore_rpc::bitcoincore_rpc_json::AddressType,
+            bitcoind::bitcoincore_rpc::RpcApi,
+        },
+        electrum_client::ElectrumApi,
+        std::time::Duration,
+        tokio::sync::OnceCell,
+    };
 
     lazy_static! {
         static ref BITCOIND: BitcoinD = {
             let bitcoind_exe = env::var("BITCOIND_EXE")
                 .ok()
-                .or(bitcoind::downloaded_exe_path().ok())
+                .or_else(|| bitcoind::downloaded_exe_path().ok())
                 .expect(
-                    "you should provide env var BITCOIND_EXE or specifiy a bitcoind version feature",
+                    "you need to provide an env var BITCOIND_EXE or specify a bitcoind version feature",
                 );
             let conf = bitcoind::Conf::default();
             BitcoinD::with_conf(bitcoind_exe, &conf).unwrap()
@@ -223,9 +228,9 @@ mod test {
         static ref ELECTRSD: ElectrsD = {
             let electrs_exe = env::var("ELECTRS_EXE")
                 .ok()
-                .or(electrsd::downloaded_exe_path())
+                .or_else(electrsd::downloaded_exe_path)
                 .expect(
-                    "you should provide env var ELECTRS_EXE or specifiy a electrsd version feature",
+                    "you need to provide env var ELECTRS_EXE or specify an electrsd version feature",
                 );
             let mut conf = electrsd::Conf::default();
             conf.http_enabled = true;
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -242,4 +245,193 @@ mod test {
...
+            .unwrap();
+        let txid_at_block_index_async = async_client
+            .get_txid_at_block_index(&block_hash, 2)
+            .await
+            .unwrap();
+        assert_eq!(txid_at_block_index, expected_txid_at_block_index);
+        assert_eq!(txid_at_block_index, txid_at_block_index_async);
+
+        // Test `get_scripthash_txs`
+        let script = Script::new();
```
tnull, line 421: While the empty script works for testing purposes here, this is quite slow since the calls return a lot of data. Suggestions for a really rare script are welcome to speed this up.

tnull:
I now covered all feasible API methods. Testing the feature parity of `get_fee_estimates` is not really feasible since the results vary quite a bit. Also note above comment regarding `scripthash_txs`.

tnull:
Squashed fixups.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ bitcoin = { version = "0.28.1", features = ["use-serde"] }
 log = "^0.4"
 ureq = { version = "~2.2.0", features = ["json"], optional = true }
 reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
+tokio = { version = "1.20.1", features = ["full"] }
```
notmandatory, line 24: Since tokio is only needed for testing the async version can we make this optional? or if only needed for testing move it to `[dev-dependencies]`.

tnull: Ah, you're totally right. Moved it to `dev-dependencies` with 8caa69b.

vladimirfomene (approved):
Tested ACK.

notmandatory:
Tests look good but I'd like to see if we can run the tests against a local regtest bitcoind and esplora rather than hitting the blockstream server URL. I'm going to give it a try using the `electrsd` crate, feel free to give it a try also, and/or when I get it working I'll show you how to do it.

tnull:
> @tnull, @vladimirfomene has the local regtest setup working in [vladimirfomene@829b2fc](https://github.com/vladimirfomene/rust-esplora-client/commit/829b2fce699ab6acebbe6ed54d22ca00737a4b18), I'd like to get his changes in as a setup function that all the tests can share. Then can we put each of your tests in it's own test function ? I know each test will be very small but it's a good idea for unit tests to only test a single thing, it will make it easier to read and to see if a method call fails which one it is.

Sounds good, will add the setup function and break the test code down into individual methods!

vladimirfomene:
@tnull, you can find the updated code here: https://github.com/vladimirfomene/rust-esplora-client/tree/add-testing

vladimirfomene:
@tnull, will it help if I refactor my code into a setup function and write a couple of test functions?

tnull:
> @tnull, will it help if I refactor my code into a setup function and write a couple of test functions?

Nah, don't worry. I already started making the changes, will probably push an update in the next 1-2 days.

tnull:
Now split the tests and updated to run against local `ElectrsD`/`BitcoinD` instances.

tnull:
> I now covered all feasible API methods. Testing the feature parity of `get_fee_estimates` is not really feasible since the results vary quite a bit.

Also added basic functionality testing of `get_fee_estimates` now.

afilini:
(needs rebase because #14 changed the CI)

src/lib.rs
```diff
@@ -198,6 +198,98 @@ impl_error!(bitcoin::hashes::hex::Error, Hex, Error);
...
+        let blocking_client = builder.build_blocking().unwrap();
+
+        let builder_async = Builder::new(&format!("http://{}", esplora_url));
+        let async_client = builder_async.build_async().unwrap();
+
+        (blocking_client, async_client)
+    }
+
+    fn generate_blocks_and_wait(num: usize) {
+        let cur_height = BITCOIND.client.get_block_count().unwrap();
```
afilini, line 248: I would move the line where you take the `MINER` lock in here, so that you don't have to remember to do it manually every time.

The main difference would be that the lock is dropped as soon as the blocks are generated, but I don't think any of your tests have explicit assertions on things like the block height, so I think it's fine if another tests generates blocks while the assertions are running

src/lib.rs
```diff
@@ -198,6 +198,98 @@ impl_error!(bitcoin::hashes::hex::Error, Hex, Error);
...
+    };
+    use electrum_client::ElectrumApi;
+    use lazy_static::lazy_static;
+    use std::sync::{Mutex, Once};
+    use std::time::Duration;
+
+    lazy_static! {
+        static ref BITCOIND: BitcoinD = {
+            let bitcoind_exe =
+                bitcoind::downloaded_exe_path().expect("bitcoind version feature must be enabled");
```
afilini, line 214: Normally we read an env variable for the bitcoind/electrsd binaries and only if that's not set we try to use the embedded one. I personally use this on nixos because the binaries downloaded by the crate don't work right out of the box, and it's just easier to have my own bitcoind/electrsd and use them

tnull:
> (needs rebase because #14 changed the CI)

Rebased on master, squashed commits, and included a minor clippy warning fix.

src/lib.rs
afilini: I think they are called `BITCOIND_EXE` and `ELECTRSD_EXE`in bdk

src/lib.rs
tnull: No, in fact I had that variant before and moved the lock back into the individual methods after encountering some race conditions from the parallel tests (namely it couldn't be asserted that the results returned from the blocking/async clients would be equal, when other tests happen to create blocks in-between the calls).

src/lib.rs
tnull: Done in https://github.com/bitcoindevkit/rust-esplora-client/pull/10/commits/7e61f2c44e7957afffd74b2c2f18b2fdf89274bc

tnull: I now adopted the somewhat inconsistent `ELECTRS_EXE` naming from BDK (vs. `ELECTRSD_EXE`) to make it more convenient.

tnull:
Wow, clippy can be a pain 😁 . But passes CI now, squashed the fixups.

vladimirfomene (commented):
Tested ACK

src/lib.rs
```diff
@@ -198,6 +198,120 @@ impl_error!(bitcoin::hashes::hex::Error, Hex, Error);
...
+        tokio::sync::OnceCell,
+    };
+
+    lazy_static! {
+        static ref BITCOIND: BitcoinD = {
+            let bitcoind_exe = env::var("BITCOIND_EXE")
+                .ok()
+                .or_else(|| bitcoind::downloaded_exe_path().ok())
+                .expect(
+                    "you should provide env var BITCOIND_EXE or specifiy a bitcoind version feature",
```
notmandatory, line 223: nit, spelling typo
```suggestion
                    "you should provide env var BITCOIND_EXE or specify a bitcoind version feature",
```

src/lib.rs
```diff
@@ -198,6 +198,120 @@ impl_error!(bitcoin::hashes::hex::Error, Hex, Error);
...
+                );
+            let conf = bitcoind::Conf::default();
+            BitcoinD::with_conf(bitcoind_exe, &conf).unwrap()
+        };
+        static ref ELECTRSD: ElectrsD = {
+            let electrs_exe = env::var("ELECTRS_EXE")
+                .ok()
+                .or_else(electrsd::downloaded_exe_path)
+                .expect(
+                    "you should provide env var ELECTRS_EXE or specifiy a electrsd version feature",
```
notmandatory, line 233: nit, spelling typo
```suggestion
                    "you should provide env var ELECTRS_EXE or specify a electrsd version feature",
```

notmandatory (approved):
ACK d1b58772c8a7fad089b23ea0c47139a8c59ccc27

I commented on some small spelling typos, but everything else looks great to me.

notmandatory:
If ok for everyone I'd like to get this published with Vlad on Mon or Tues so we can get a PR into `bdk` for the next release to remove the esplora client code there and use this crate.

PR #13, opened 2022-09-16
Query block hash by by block height

gcomte:
Give the possibility to query a block hash by providing its corresponding block height

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -115,22 +115,11 @@ impl AsyncClient {
 
     /// Get a [`BlockHeader`] given a particular block height.
     pub async fn get_header(&self, block_height: u32) -> Result<BlockHeader, Error> {
-        let resp = self
-            .client
-            .get(&format!("{}/block-height/{}", self.url, block_height))
-            .send()
-            .await?;
-
-        if let StatusCode::NOT_FOUND = resp.status() {
-            return Err(Error::HeaderHeightNotFound(block_height));
-        }
-        let bytes = resp.bytes().await?;
-        let hash =
-            std::str::from_utf8(&bytes).map_err(|_| Error::HeaderHeightNotFound(block_height))?;
+        let block_hash = self.get_block_hash(block_height).await?;
 
         let resp = self
             .client
-            .get(&format!("{}/block/{}/header", self.url, hash))
+            .get(&format!("{}/block/{}/header", self.url, block_hash))
             .send()
             .await?;
 
@@ -203,6 +192,23 @@ impl AsyncClient {
             .get(&format!("{}/blocks/tip/hash", self.url))
             .send()
             .await?;
+
+        Ok(BlockHash::from_str(
+            &resp.error_for_status()?.text().await?,
+        )?)
+    }
+
+    /// Get the [`BlockHash`] of a specific block height
+    pub async fn get_block_hash(&self, block_height: u32) -> Result<BlockHash, Error> {
+        let resp = self
+            .client
+            .get(&format!("{}/block-height/{}", self.url, block_height))
+            .send()
+            .await?;
+
+        if let StatusCode::NOT_FOUND = resp.status() {
+            return Err(Error::HeaderHeightNotFound(block_height));
+        }
 
         Ok(BlockHash::from_str(
             &resp.error_for_status()?.text().await?,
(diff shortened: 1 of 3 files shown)


notmandatory:
This looks good, @tnull and @vladimirfomene are also working on examples for testing, once their stuff is in you should add a similar test for this new query also.

tnull (commented):
Generally looks good. Maybe we want to reuse these methods in `get_header`, since it also gets the block hash by height?

vladimirfomene:
I believe we need to be careful when reusing some of these methods in other methods as it might result into doing more than one round trip to the server.

tnull:
> I believe we need to be careful when reusing some of these methods in other methods as it might result into doing more than one round trip to the server.

Ah, excuse the mistake, I meant `get_header`, in which two request are sent anyways currently?:

https://github.com/bitcoindevkit/rust-esplora-client/blob/590a01bd62d489a2e862bddd8077dce8c25b10db/src/blocking.rs#L130-L148

gcomte:
Imo reusing `get_block_hash` within `get_header` makes sense.

Looking at the `get_header` code, I realize that I should check specifically for HTTP Error 404, in order to return `HeaderHeightNotFound`. However, this is only done by the async client, the blocking one just directly returns the HTTP error code. What's the reasoning behind this difference?

src/async.rs
```diff
@@ -120,22 +120,11 @@ impl AsyncClient {
...
-            .send()
-            .await?;
-
-        if let StatusCode::NOT_FOUND = resp.status() {
-            return Err(Error::HeaderHeightNotFound(block_height));
-        }
-        let bytes = resp.bytes().await?;
-        let hash =
-            std::str::from_utf8(&bytes).map_err(|_| Error::HeaderHeightNotFound(block_height))?;
+        let block_hash = self.get_block_hash(block_height).await?.to_string();
```
tnull, line 123: nit: I think clippy will complain otherwise since `BlockHash` implements `Display`.
```suggestion
        let block_hash = self.get_block_hash(block_height).await?;
```

src/blocking.rs
```diff
@@ -133,23 +133,11 @@ impl BlockingClient {
...
-
-        let bytes = match resp {
-            Ok(resp) => Ok(into_bytes(resp)?),
-            Err(ureq::Error::Status(code, _)) => Err(Error::HttpResponse(code)),
-            Err(e) => Err(Error::Ureq(e)),
-        }?;
-
-        let hash =
-            std::str::from_utf8(&bytes).map_err(|_| Error::HeaderHeightNotFound(block_height))?;
+        let block_hash = self.get_block_hash(block_height)?.to_string();
```
tnull, line 136: nit: Likewise.
```suggestion
        let block_hash = self.get_block_hash(block_height)?;
```

tnull (commented):
Generally looks good, two nits.

gcomte:
> Generally looks good, two nits.

Fixed and rebased onto `master`. Clippy happy now.

tnull (approved):

vladimirfomene:
@gcomte, I think we can squash these changes into one commit.

gcomte:
Yeah, that could be done when merging, right?

vladimirfomene:
Hey @gcomte ! can you squash it here? It makes things easy for us.

gcomte:
I guess I should write some tests as well, as this PR has been merged now: https://github.com/bitcoindevkit/rust-esplora-client/pull/10

vladimirfomene:
@gcomte, can you add test to this so that I can merge it?

gcomte:
@vladimirfomene : b6d663b2fff5fcbcc13bd8093de9b8f6fb29e433

Do you want me to squash it into the other commit?

src/lib.rs
```diff
@@ -545,6 +545,15 @@ mod test {
         assert_eq!(tip_hash, tip_hash_async);
     }
 
+    #[cfg(all(feature = "blocking", any(feature = "async", feature = "async-https")))]
+    #[tokio::test]
+    async fn test_get_block_hash() {
+        let (blocking_client, async_client) = setup_clients().await;
+        let block_hash = blocking_client.get_block_hash(21).unwrap();
+        let block_hash_async = async_client.get_block_hash(21).await.unwrap();
+        assert_eq!(block_hash, block_hash_async);
```
tnull, line 554: You may want to add another check comparing the Esplora results to RPC, i.e., `BITCOIND.client.get_block_hash(21)`

gcomte: That's right. Added: https://github.com/bitcoindevkit/rust-esplora-client/commit/7c66cf1f0058feb11bff3bacad98068bfd1bbf82

tnull (approved):

vladimirfomene (approved):

PR #17, opened 2022-10-06
Allow to retrieve a block header by hash

tnull:
~~Based on #13 since it probably land soon.~~

This PR adds a method allowing to retrieve a `BlockHeader` via a given `BlockHash` and reuses it in `get_height`.

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -129,7 +129,11 @@ impl BlockingClient {
     /// Get a [`BlockHeader`] given a particular block height.
     pub fn get_header(&self, block_height: u32) -> Result<BlockHeader, Error> {
         let block_hash = self.get_block_hash(block_height)?;
+        self.get_header_by_hash(&block_hash)
+    }
 
+    /// Get a [`BlockHeader`] given a particular block hash.
+    pub fn get_header_by_hash(&self, block_hash: &BlockHash) -> Result<BlockHeader, Error> {
         let resp = self
             .agent
             .get(&format!("{}/block/{}/header", self.url, block_hash))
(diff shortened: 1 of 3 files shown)


src/blocking.rs
```diff
@@ -128,23 +128,15 @@ impl BlockingClient {
 
     /// Get a [`BlockHeader`] given a particular block height.
     pub fn get_header(&self, block_height: u32) -> Result<BlockHeader, Error> {
```
tnull, line 130: Not sure if we should rename it to `get_header_by_height`.

gcomte: +1 on removing it. If anything, an API for getting a header by block_height should be implemented by Esplora, the library should just reflect what the Esplora API offers.

notmandatory: I agree `get_header` should be removed since it doesn't reflect an existing Esplora API. But since we have at least one user (`bdk`) can you just [deprecate](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute) it for now?

notmandatory:
Overall looks good to me, but for the new functions would be good to have at least some simple tests added in lib.rs.

src/blocking.rs
tnull: Deprecated with https://github.com/bitcoindevkit/rust-esplora-client/pull/17/commits/68b9d2b15366b04c28bc9dbeac1c2efb321fd63b.

tnull:
> Overall looks good to me, but for the new functions would be good to have at least some simple tests added in lib.rs.

Added a simple test case for `get_block_by_height` and removed the test for the now-deprecated `get_height` with https://github.com/bitcoindevkit/rust-esplora-client/pull/17/commits/e01061177bbb32a170c977f2c62ddf8178536311.

vladimirfomene (approved):

vladimirfomene (approved):

tnull:
Rebased on master after #13 landed.

PR #22, opened 2022-10-25
Fix deprecation version of `get_header`

tnull:
In #17, we deprecated `get_header` and assumed the next version would be 0.1.2, but it turned out to be 0.2.


vladimirfomene (approved):
Code review ACK. This makes sense to me.

PR #28, opened 2022-11-04
Add `get_merkle_block` API call

tnull:
We already have the `get_merkle_proof` call, but this one returns a `bitcoin::MerkleBlock`, which has additional data and is more often than not more useful than Electrum's counterpart.

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -17,7 +17,7 @@ use std::str::FromStr;
 use bitcoin::consensus::{deserialize, serialize};
 use bitcoin::hashes::hex::{FromHex, ToHex};
 use bitcoin::hashes::{sha256, Hash};
-use bitcoin::{Block, BlockHash, BlockHeader, Script, Transaction, Txid};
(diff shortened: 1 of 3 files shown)


tnull:
Huh, I think the clippy error is unrelated to my change?

notmandatory:
> Huh, I think the clippy error is unrelated to my change and the warning seems to only be added by the just-released 1.65.0 (see https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err). Not sure how to go about it? Should I fix it here, or do we want to do that in another PR?

I think this is OK to fix in a different PR, for now you can allow the warning where needed with:
```rust
#[allow(clippy::result_large_err)]
```

tnull:
> I think this is OK to fix in a different PR, for now you can allow the warning where needed with:

Alright, I now set the attribute crate-wide and squashed again.

notmandatory (approved):
utACK e40583af13bd4ce812e91476ffa489553b4a26de

vladimirfomene (approved):
Tested ACK

PR #27, opened 2022-11-01
Add /blocks API

LLFourn:
Note that all the fields have been added yet. Just some useful ones that seemed important for bdk.

PR #29, opened 2022-11-26
Add `get_recent_blocks` api

rajarshimaitra:
This adds a new api `get_recent_blocks` that returns the last 10 block heights and hash, with an optional starting height.

Following https://github.com/Blockstream/esplora/blob/master/API.md#get-blocksstart_height

This is required in bdk_core esplora syncing..

Dependent PR : <TODO>

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -79,6 +79,12 @@ pub struct BlockTime {
     pub height: u32,
 }
 
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct Block {
+    pub id: BlockHash,
+    pub height: u32,
+}
+
 impl Tx {
     pub fn to_tx(&self) -> Transaction {
         Transaction {
--- a/src/async.rs
+++ b/src/async.rs
@@ -136,6 +136,20 @@ impl AsyncClient {
         Ok(header)
     }
 
+    /// Returns the 10 newest blocks starting at the tip or at start_height if specified.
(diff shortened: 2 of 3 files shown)


src/api.rs
```diff
@@ -79,6 +79,12 @@ pub struct BlockTime {
     pub height: u32,
 }
 
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct Block {
```
tnull, line 83: Can we name this differently so that it doesn't shadow `bitcoin::Block`?

tnull (commented):
Mh, I believe this is a duplicate of #27?

rajarshimaitra:
> Mh, I believe this is a duplicate of https://github.com/bitcoindevkit/rust-esplora-client/pull/27?

I saw #27 after opening this one.. Although waiting on https://github.com/bitcoindevkit/rust-esplora-client/pull/27#discussion_r1032889211 to use it in bdk_core..

I am open to reject this one in favor of #27. That seems more detailed..

src/api.rs
rajarshimaitra: ACK.. Will do if we finalize on this..

vladimirfomene:
@rajarshimaitra, can you rebase this changes on the current master branch?

LLFourn:
I fixed up #27 let's use that.

PR #31, opened 2022-12-15
Run tests in a single thread, plus minor ci cleanups

notmandatory:
I changed the CI tests to run in a single thread (one at a time) which should improve test reliability. I also added names to all the test tasks, and remove redundant clippy checks.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -26,13 +26,13 @@ jobs:
       run: cargo fmt -- --check --verbose
     - name: Build
       run: cargo build --verbose --all-features
-    - name: Test
-      run: cargo test --verbose --all-features
-    - run: cargo check --verbose --no-default-features --features=blocking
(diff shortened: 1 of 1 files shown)


tnull:
Ah, if it should be single-threaded going forward, I believe the setup logic could be simplified a bit too, e.g., the premine probably doesn't need to happen synchronized via `tokio::sync::OnceCell` (but can, of course):

https://github.com/bitcoindevkit/rust-esplora-client/blob/3030d2b099330ea6f0752ba89acc294a48ac3946/src/lib.rs#L247

notmandatory:
I think we still need that setup logic since the single thread is just the CI system running tests one at a time.. within each tests (the async ones at least) there could be multiple threads.

vladimirfomene (approved):
LGTM

oleonardolima:
I wonder what's wrong with these coverage reports by coveralls being sent now 🤔.

PR #40, opened 2023-03-04
Disable default features for rust-bitcoin

benthecarman:
This shouldnt be needed and will let this be used better in wasm contexts because often you need the no-std version of libraries due to system time.


benthecarman:
nvm saw #39 was already created

PR #44, opened 2023-03-25
fix: expose TLS backend choice in features

### LINKED ISSUE #43: Expose TLS implementation choice via features
Some projects (e.g. @fedimint, https://github.com/fedimint/fedimint/issues/2032#issue-1640445064) prefer `rustls` over OpenSSL, it would be nice if the choice could be propagated through.

elsirion:
I left the default `async-https` feature in to make this a non-breaking change. Could be phased out though imo, just make one of the new features a default feature instead.

Fixes #43


vladimirfomene (approved):
This looks good to me. Thanks for the contribution! Will discuss with team whether it makes sense to phase out the `async-https` feature.

Cargo.toml
```diff
@@ -38,3 +38,5 @@ default = ["blocking", "async", "async-https"]
 blocking = ["ureq", "ureq/socks-proxy"]
 async = ["reqwest", "reqwest/socks"]
 async-https = ["async", "reqwest/default-tls"]
+async-https-native = ["async", "reqwest/native-tls"]
+async-https-rustls = ["async", "reqwest/rustls"]
```
notmandatory, line 42: I think we're supposed to use the feature `request/rustls-tls` which defaults to using [webpki-roots](https://github.com/rustls/webpki-roots), or do you want manually set your root certs? In that case we'd need `request/rustls-tls-manual-roots`.

notmandatory: Or can add both options:
```suggestion
async-https-rustls = ["async", "reqwest/rustls-tls"]
async-https-rustls-manual-roots = ["async", "reqwest/rustls-tls-manual-roots"]
```

elsirion:
What's the status on this? Do I need to change anything?

vladimirfomene:
Sorry for the delay on this. This is good. I'm merging it.

PR #46, opened 2023-04-25
bugfix: `GET /tx/:txid/status` is not `Option`

MaxFangX:
See the [server-side handler](https://github.com/Blockstream/electrs/blob/adedee15f1fe460398a7045b292604df2161adc0/src/rest.rs#L941) - unlike the other `tx` endpoints, this one doesn't first check if the tx is known.

Also added a test that confirms this behavior.


vladimirfomene (approved):
Reviewed ACK

PR #51, opened 2023-06-10
Upgrade bitcoin to v0.30.0

tcharding:
Upgrade `rust-bitcoin` to v0.30.3, including:

- bitcoin
- electrsd
- electrum-client

Also includes addition of a dependency on `bitcoin-internals` for the `DisplayHex` trait. This is a temporary fix while we wait for release of the new [`github.com/rust-bitcoin/hex-conservative`](https://github.com/rust-bitcoin/hex-conservative) crate.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -48,6 +48,9 @@ jobs:
       run: rustup component add clippy
     - name: Update toolchain
       run: rustup update
+    - name: pin dependencies
+      if: ${{ matrix.rust.version }} == 1.57.0
+      run: cargo update -p log --precise 0.4.18 && cargo update -p rustls@0.21.2 --precise 0.21.1
     - name: Build
       run: cargo build --features ${{ matrix.features }} --no-default-features
     - name: Clippy
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -49,8 +49,8 @@ jobs:
     - name: Update toolchain
       run: rustup update
     - name: pin dependencies
-      if: ${{ matrix.rust.version }} == 1.57.0
-      run: cargo update -p log --precise 0.4.18 && cargo update -p rustls@0.21.2 --precise 0.21.1
+      if: matrix.rust.version == '1.57.0'
+      run: cargo update -p log --precise 0.4.18 && cargo update -p rustls:0.21.2 --precise 0.21.1 && cargo update -p time:0.3.15 --precise 0.3.13
     - name: Build
       run: cargo build --features ${{ matrix.features }} --no-default-features
     - name: Clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,9 @@ path = "src/lib.rs"
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-bitcoin = { version = "0.29.1", features = ["serde"], default-features = false }
+bitcoin = { version = "0.30.0", features = ["serde", "std"], default-features = false }
+# Temporary dependency on internals until the rust-bitcoin devs release the hex-conservative crate.
+bitcoin-internals = { version = "0.1.0", features = ["alloc"] }
 log = "^0.4"
 ureq = { version = "2.5.0", features = ["json"], optional = true }
 reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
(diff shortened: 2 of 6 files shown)


vladimirfomene (commented):
Code LGTM. I was wondering why we suddenly need a `std` feature for `rust-bitcoin`. Will having this feature stop using the client in a `no-std` environment? I think you should pin the `log` dependency to pass CI MSRV test. See: https://github.com/bitcoindevkit/bdk/commit/fa54a2e3a5096b263388b9f729ba1cb42574072e

tcharding:
> Code LGTM. 

Thanks for the review.

> I was wondering why we suddenly need a `std` feature for `rust-bitcoin`. Will having this feature stop using the client in a `no-std` environment?

I didn't see `#[no_std]` and also in `lib.rs` there are a bunch of `std` imports so I assumed this crate was not trying to be no-std compatible.

> I think you should pin the `log` dependency to pass CI MSRV test. See: [bitcoindevkit/bdk@fa54a2e](https://github.com/bitcoindevkit/bdk/commit/fa54a2e3a5096b263388b9f729ba1cb42574072e)

Cool, thanks - I'll have a go at that.

tcharding:
I can re-order the commits if this passes CI, I'm not totally sure of the syntax used in github actions to achieve pinning.

Cargo.toml
```diff
@@ -17,16 +17,18 @@ path = "src/lib.rs"
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-bitcoin = { version = "0.29.1", features = ["serde"], default-features = false }
+bitcoin = { version = "0.30.0", features = ["serde", "std"], default-features = false }
+# Temporary dependency on internals until the rust-bitcoin devs get their shit together and release the hex-conservative crate.
```
notmandatory, line 21: Go easy on the poor devs! ;-) 
```suggestion
# Temporary dependency on internals until the rust-bitcoin devs release the hex-conservative crate.
```

notmandatory:
Looks like build is breaking because rustls changed their MSRV from 1.57.0 (in 0.21.1) to 1.60.0 (in 0.21.2). Yet again I wish crates wouldn't changes their MSRV in patch releases. Can we pin this the same way you did `log`?

tcharding:
I re-ordered the commits to put the pinning first. (And I removed attack on myself since others won't necessarily know it was me who both wrote the comment and caused the problem :)

.github/workflows/cont_integration.yml
```diff
@@ -48,6 +48,9 @@ jobs:
       run: rustup component add clippy
     - name: Update toolchain
       run: rustup update
+    - name: pin dependencies
+      if: ${{ matrix.rust.version }} == 1.57.0
+      run: cargo update -p log --precise 0.4.18 && cargo update --p rustls --precise 0.21.1
```
notmandatory, line 53: Little typo here:
```suggestion
      run: cargo update -p log --precise 0.4.18 && cargo update -p rustls --precise 0.21.1
```

LLFourn:
> Looks like build is breaking because rustls changed their MSRV from 1.57.0 (in 0.21.1) to 1.60.0 (in 0.21.2). Yet again I wish crates wouldn't changes their MSRV in patch releases. Can we pin this the same way you did `log`?
> 
> https://github.com/rustls/rustls/blob/v/0.21.1/rustls/Cargo.toml

The cargo documentation states that changing the rust version is only a "possibly breaking change": https://doc.rust-lang.org/cargo/reference/semver.html#env-new-rust

...where the possibility is 100% if you use an older version of rust.

tcharding:
Why not just pin in CI for MSRV build and add a comment in the README to tell folks wishing to build with MSRV? (Of note this PR does not include a comment in the README, my bad - can amend).

Just because `rustls` did the wrong thing we don't want to punish users of `bdk` that are not using MSRV by forcing them to use the old version, right?

tcharding:
Added notes on pinning to readme.

LLFourn:
Since this PR is a breaking change anyway why not just increase the MSRV for this crate?

vladimirfomene:
@tcharding, the CI is still failing. If you can fix it. I will merge it asap

tcharding:
> Writing in the README is code that has to be maintained and can't be tested. It's better not to.

Fair point. Will remove.

tcharding:
> Since this PR is a breaking change anyway why not just increase the MSRV for this crate?

I don't know how you guys chose your MSRV so I don't want to bump it. More generally I don't want to do it just because `rustls` did it.

tcharding:
> CI is still failing. If you can fix it, I will merge asap

The problem was that we have two versions of `rustls` in the dependency graph. One, that needs pinning, comes from hyper stack. The other comes from `electrum-client`.

vladimirfomene:
@tcharding do you mind creating a PR on the `electrum-client` repo for this update?

tcharding:
I added a patch to remind us to fix the pinning introduced by this PR after https://github.com/bitcoindevkit/rust-electrum-client/pull/110 merges. I'm not across your release cycle so please instruct me if you'd like it removed or anything else different.

Thanks.

tcharding:
Thanks for catching me @ vladimirfomene :)

vladimirfomene (commented):
@tcharding thanks for these changes! Re-ACK d48ac225d3be20ba980391920eb17441a8d91697

vladimirfomene (approved):

PR #52, opened 2023-06-13
Remove unused dev-dependencies

tcharding:
The two dependencies `zip` and `base64ct` are not used, remove them.

If this was meant as a way to fix the version of these libs we can always pin in CI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,10 +28,6 @@ tokio = { version = "1.20.1", features = ["full"] }
 electrsd = { version = "0.22.0", features = ["legacy", "esplora_a33e97e1", "bitcoind_22_0"] }
 electrum-client = "0.12.0"
 lazy_static = "1.4.0"
-# zip versions after 0.6.3 don't work with our MSRV 1.57.0
-zip = "=0.6.3"
-# base64ct versions at 1.6.0 and higher have MSRV 1.60.0
-base64ct = "<1.6.0"
 
 [features]
 default = ["blocking", "async", "async-https"]


vladimirfomene:
@tcharding yes we pinned those dependencies to fix MSRV of some of our dependencies so that they come down to 1.57.0 which is BDK's current MSRV. How will pinning in CI work? Will pinning in CI ensure that `rust-esplora-client` maintain MSRV of 1.57.0 during release?

notmandatory:
Both `zip` and `base64ct` are needed run the tests but we should be able to only pin them in [dev-dependencies].

notmandatory:
It would be nice to get these versions pinned in `electrsd`, issue and discussion here: https://github.com/RCasatta/electrsd/issues/56

tcharding:
In `rust-bitcoin` we run a shell script in CI for test. And in that we pin dependencies if we are running the script while using the MSRV toolchain

You can check out https://github.com/rust-bitcoin/rust-bitcoin/blob/master/bitcoin/contrib/test.sh if interested, and here is the relevant snippet:
```
# Pin dependencies as required if we are using MSRV toolchain.
if cargo --version | grep "1\.48"; then
    # 1.0.157 uses syn 2.0 which requires edition 2018
    cargo update -p serde --precise 1.0.156
fi
```

tcharding:
I'm going to close this PR because I'm unlikely to get to changing your CI, and I don't really want to mess with the CI of another project, seems a bit rude :)

PR #55, opened 2023-07-30
Refactor to use BP stack instead of rust-bitcoin

dr-orlovsky:
This makes library much more lightweight, helps to stabilize API and avoid major refactorings with each new `rust-bitcoin` release.


dr-orlovsky:
Sorry, GitHub opened PR in the wrong repo! It was intended to be against the fork In https://github.com/BP-WG/bp-esplora-client

PR #59, opened 2023-10-08
fix(wallet_esplora): Gracefully handle 429s errors

realeinherjar:
Closes https://github.com/bitcoindevkit/bdk/issues/1120

## Details

- `BlockingClient`:
  - Add an extra check inside `scripthash_txs` for 429 Status codes.
    If detected, hard-coded sleep for 500ms using `std::thread::sleep` (so it is also thread-safe).
- `AsyncClient` (little more convoluted than `BlockingClient`):
  - I had to bring `tokio` and `async-recursion` into the `dependencies`.
  - Inside the `scripthash_txs` we are checking for 429 Status codes.
    If detected, hard-coded sleep for 500ms using `tokio::time::sleep`.
    Since this is an "async recursion" the macro [`async_recursion`](https://docs.rs/async-recursion) handles that gracefully without having to deal with:

    ```rust
    # https://rust-lang.github.io/async-book/07_workarounds/04_recursion.html
    use futures::future::{BoxFuture, FutureExt};

    fn recursive() -> BoxFuture<'static, ()> {
        async move {
            recursive().await;
            recursive().await;
        }.boxed()
    }
    ```

    If we annotate the offending recursive function with `#[async_recursion]`,
    it automatically convert an async function to one returning a boxed `Future`.

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -15,6 +15,7 @@ use std::collections::HashMap;
 use std::io;
 use std::io::Read;
 use std::str::FromStr;
+use std::thread::sleep;
 use std::time::Duration;
 
 #[allow(unused_imports)]
@@ -380,7 +381,26 @@ impl BlockingClient {
             ),
             None => format!("{}/scripthash/{:x}/txs", self.url, script_hash),
         };
-        Ok(self.agent.get(&url).call()?.into_json()?)
+        let resp = self.agent.get(&url).call(); //?.into_json();
+        match resp {
+            Ok(resp) => {
+                let resp_json = resp.into_json()?;
+                Ok(resp_json)
+            }
+            Err(ureq::Error::Status(code, resp)) => {
+                if is_status_too_many_requests(code) {
+                    eprintln!("Too many requests, sleeping for 500 ms");
+                    sleep(Duration::from_millis(500));
+                    self.scripthash_txs(script, last_seen)
+                } else {
+                    Err(Error::HttpResponse {
+                        status: code,
+                        message: resp.into_string()?,
+                    })
+                }
+            }
+            Err(e) => Err(Error::Ureq(e)),
+        }
     }
 
     /// Gets some recent block summaries starting at the tip or at `height` if provided.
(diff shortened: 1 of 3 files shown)


evanlinjin (changes requested):
Aproach NACK

Thank you for taking a shot at this.

Why 500ms? What if the server expects us to wait for 600ms? I would prefer if we checked the `Retry-After` header. If that does not exist, then have an exponential back-off delay as a fallback.

Do we need the async recursion? Is there a way to do it with a loop?

I also think introducing `tokio` as an async dependency is a big ask. Would it be trivial to implement a simple `Future` using [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html)?

realeinherjar:
Thanks for the feedback. It is helpful. I will try to modify the approach. 👍🏻

LLFourn:
> Aproach NACK
> 
> Thank you for taking a shot at this.
> 
> Why 500ms? What if the server expects us to wait for 600ms? I would prefer if we checked the `Retry-After` header. If that does not exist, then have an exponential back-off delay as a fallback.
> 
> Do we need the async recursion? Is there a way to do it with a loop?
> 
> I also think introducing `tokio` as an async dependency is a big ask. Would it be trivial to implement a simple `Future` using [`Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html)?

Not really trivial and using `std::time::Instant` breaks WASM probably. We'd need to do some work to have the timer work on WASM: See: https://github.com/whizsid/wasmtimer-rs

I would avoid sleeping and just lower the number of batch requests in the next round by the number of 429s you get.

PS are we testing this crate on WASM? If not PR fixing that before this one would be good.

src/blocking.rs
```diff
@@ -380,7 +381,26 @@ impl BlockingClient {
...
+        let resp = self.agent.get(&url).call(); //?.into_json();
+        match resp {
+            Ok(resp) => {
+                let resp_json = resp.into_json()?;
+                Ok(resp_json)
+            }
+            Err(ureq::Error::Status(code, resp)) => {
+                if is_status_too_many_requests(code) {
+                    eprintln!("Too many requests, sleeping for 500 ms");
+                    sleep(Duration::from_millis(500));
```
benthecarman, line 393: this will cause a panic in wasm, might need to pass in a function that sleeps. for wasm sleep needs to be async too

realeinherjar:
I will try another approach in the `bdk_esplora` crate and not here.

PR #62, opened 2023-10-25
ci: test WASM

realeinherjar:
Since #59 mentioned that wasm would break under certain circumstances, we have no idea when it will break.
Hence, this PR adds WASM checks in the CI.

From @LLFourn in #59:
> PS are we testing this crate on WASM? If not PR fixing that before this one would be good.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,17 +17,28 @@ path = "src/lib.rs"
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
-bitcoin = { version = "0.30.0", features = ["serde", "std"], default-features = false }
+bitcoin = { version = "0.30.0", features = [
+  "serde",
+  "std",
+], default-features = false }
 # Temporary dependency on internals until the rust-bitcoin devs release the hex-conservative crate.
 bitcoin-internals = { version = "0.1.0", features = ["alloc"] }
 log = "^0.4"
 ureq = { version = "2.5.0", features = ["json"], optional = true }
-reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
+reqwest = { version = "0.11", optional = true, default-features = false, features = [
+  "json",
+] }
+# WASM specific fix
+getrandom = { version = "0.2", features = ["js"] }
 
 [dev-dependencies]
 serde_json = "1.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -17,17 +17,28 @@ path = "src/lib.rs"
...
+  "std",
+], default-features = false }
 # Temporary dependency on internals until the rust-bitcoin devs release the hex-conservative crate.
 bitcoin-internals = { version = "0.1.0", features = ["alloc"] }
 log = "^0.4"
 ureq = { version = "2.5.0", features = ["json"], optional = true }
-reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
+reqwest = { version = "0.11", optional = true, default-features = false, features = [
+  "json",
+] }
```
realeinherjar, line 30: auto formatting

Cargo.toml
```diff
@@ -17,17 +17,28 @@ path = "src/lib.rs"
...
 
 [dev-dependencies]
 serde_json = "1.0"
 tokio = { version = "1.20.1", features = ["full"] }
-electrsd = { version = "0.24.0", features = ["legacy", "esplora_a33e97e1", "bitcoind_22_0"] }
+electrsd = { version = "0.24.0", features = [
+  "legacy",
+  "esplora_a33e97e1",
+  "bitcoind_22_0",
+] }
```
realeinherjar, line 41: auto formatting

Cargo.toml
```diff
@@ -17,17 +17,28 @@ path = "src/lib.rs"
...
 # Temporary dependency on internals until the rust-bitcoin devs release the hex-conservative crate.
 bitcoin-internals = { version = "0.1.0", features = ["alloc"] }
 log = "^0.4"
 ureq = { version = "2.5.0", features = ["json"], optional = true }
-reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
+reqwest = { version = "0.11", optional = true, default-features = false, features = [
+  "json",
+] }
+# WASM specific fix
+getrandom = { version = "0.2", features = ["js"] }
```
realeinherjar, line 32: Only real change in `Cargo.toml`

realeinherjar:
MSRV is broken. I will work on a fix in a new PR and rebase this one.

PR #67, opened 2023-12-08
wip(feat): add new `AsyncAnonymizedClient` using `arti-hyper`

oleonardolima:
- add new async client, `AsyncAnonymizedClient`, that uses `arti-hyper`, and `arti-client` to connect and do requests over the Tor network.


notmandatory:
Please rebase to pick up changes in #69 that fix CI.

oleonardolima:
Most of the implementation is done, and it currently works on rust 1.70, also good to review and get inputs from it's being done.

But it does not work with our current MSRV, how do you usually go forward? just pin some of the dependencies?

cc @notmandatory @vladimirfomene

notmandatory:
@oleonardolima can your rebase and rework this now that #75 is merged? hopefully not too big a change.

Cargo.toml
```diff
@@ -18,27 +18,44 @@ path = "src/lib.rs"
...
 blocking-https-native = ["blocking", "minreq/https-native"]
 blocking-https-bundled = ["blocking", "minreq/https-bundled"]
+
 async = ["reqwest", "reqwest/socks"]
 async-https = ["async", "reqwest/default-tls"]
 async-https-native = ["async", "reqwest/native-tls"]
 async-https-rustls = ["async", "reqwest/rustls-tls"]
 async-https-rustls-manual-roots = ["async", "reqwest/rustls-tls-manual-roots"]
+
+# TODO: (@leonardo) Should I rename it to async-anonymized ?
```
notmandatory, line 58: how about `async-tor` instead of `async-anonymized` ? It's shorter and most users know what `tor` is but may not know about `arti`.

oleonardolima:
I'm closing this PR.

It's become stale, and I haven't worked on this for a while. I don't think this is still the right approach, as we are now moving towards using `bitreq`.

Once we have migrated to `bitreq` and it has sans-io support, we'll probably only need a simple example on how to use `arti-client` to use the `esplora-client` through Tor, or any other transport layer.

PR #69, opened 2023-12-29
ci: change MSRV to 1.63.0

notmandatory:


notmandatory:
Also rustc 1.63 is the version shipped with the current debian stable (bookworm): https://packages.debian.org/stable/rust/rustc

PR #72, opened 2024-01-18
fix: bump `h2` version to patched one, while reqwest has not been

oleonardolima:
patched yet

- check: https://rustsec.org/advisories/RUSTSEC-2024-0003


oleonardolima:
@notmandatory @vladimirfomene As we don't have the `Cargo.lock` file I wasn't sure about the best way to update/bump it.

notmandatory:
Can we bump the dependency that brings in h2 to get a patched version rather than just bumping h2?

oleonardolima:
> Can we bump the dependency that brings in h2 to get a patched version rather than just bumping h2?

@notmandatory Actually we can close this PR, I did deeper research, and the `request` crate does have a specific `h2` version (they have `^0.3.14`), and neither it's being pinned nor are we providing a `Cargo.lock` file, so it's fine as it is.

PR #60, opened 2023-10-16
Upgrade bitcoin

tcharding:
Upgrade dependencies to use the latest `rust-bitcoin v0.31`.
    
While we are at it, bump the crate version ready for release.

PR #74, opened 2024-01-27
update bitcoin to 0.31

getong:
This is a fix of #60 , which update rust to 1.67.
It fix the ci error.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "esplora-client"
-version = "0.6.0"
+version = "0.7.0"
 edition = "2018"
 authors = ["Alekos Filini <alekos.filini@gmail.com>"]
 license = "MIT"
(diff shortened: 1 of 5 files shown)


getong:
> I don't think we want to bump the MSRV on this crate to 1.67.0. We will however need to find another way forward, e.g., by carefully pinning the respective offending dependency in CI (I think in this case it would be safe as it's just a downstream dependency of our `electrsd` test dependency I believe), or by introducing a script to download and run `bitcoind`/`electrs` as we did in LDK.



> I don't think we want to bump the MSRV on this crate to 1.67.0. We will however need to find another way forward, e.g., by carefully pinning the respective offending dependency in CI (I think in this case it would be safe as it's just a downstream dependency of our `electrsd` test dependency I believe), or by introducing a script to download and run `bitcoind`/`electrs` as we did in LDK.

I am afraid not.

electrum-client use bitcoin 0.31 from version 0.19.
electrsd use electrum-client 0.19 from 0.27, so electrsd cannot be downstream here.

getong:
I revert the code, but this error seems that it might be jobserver,  I make a issue of it. 
rust-lang/jobserver-rs#69

tcharding:
Hey mate, I pushed up #79 to do the upgrade which is getting past CI now.

PR #75, opened 2024-01-31
Replace `ureq` with `minreq` for `blocking`

tnull:
Taking this over from #57.

So far just a few minor adjustments on top of #57. Tested that syncing LDK works with the updated blocking client (in addition to the tests here).

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -22,6 +22,10 @@ jobs:
         features:
           - default
           - blocking
+          - blocking-https
+          - blocking-https-rustls
+          - blocking-https-native
+          - blocking-https-bundled
           - async
           - async-https
           - async-https-native
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -57,7 +57,6 @@ jobs:
       run: |
         cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
         cargo update -p time --precise "0.3.20"
-        cargo update -p jobserver --precise "0.1.26"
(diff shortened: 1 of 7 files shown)


notmandatory:
You may need to pin some dependencies to get CI to build, see https://github.com/bitcoindevkit/bdk/blob/f099b42005b5310cee51fa2eaa29f94d76753bd7/.github/workflows/cont_integration.yml#L30C7-L36C49

notmandatory:
I'm much less concerned about pinning dev only dependencies so I don't see a reason to redo the CI to make progress on this PR. For BDK there's a PR to change the CI to use NIX which would also remove the auto-downloaded bitcoind / electrsd issue that required these broken for msrv dev dependencies (https://github.com/bitcoindevkit/bdk/pull/1257). Once we get that PR into BDK we should be able to do something similar for this project with @storopoli 's help.

storopoli:
Yeah count me in. Propagating Nix CI PRs throughout BDK's ecosystem should be a breeze after https://github.com/bitcoindevkit/bdk/pull/1257

notmandatory:
Rather than mixing the discussion and fixes for the MSRV issue into this PR I've create a stand-alone PR #76 for it.

notmandatory:
I merged #79 so this PR should be re-based again.

notmandatory:
I had to push another commit to fix the CI. Apparently the latest version of `jobserver` has an MSRV of 1.63 so we do not need to pin it anymore (and our old pinned version 1.26 does not work).

notmandatory (approved):
ACK 6a598b270cd64bdadde69126db1a2b200ce96231

PR #76, opened 2024-02-03
Bump electrsd to 0.26 + bitcoind_25_0 and fix msrv errors

notmandatory:
This PR bumps electrsd to 0.26 and replaces temporary dependency on `bitcoin-internals` with `hex-conservative` as done in #75. Fixed CI MSRV issues with dev dependencies blocking #75 and #74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,5 +18,5 @@ This library should compile with any combination of features with Rust 1.63.0.
 To build with the MSRV you will need to pin dependencies as follows:
 
 ```shell
-cargo update -p home --precise 0.5.5
+cargo update -p jobserver --precise "0.1.26"
 ```
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -44,7 +44,7 @@ jobs:
         # For more info, see:
         # - https://github.com/bitcoindevkit/bdk/issues/696
         # - https://github.com/bitcoindevkit/bdk/pull/748#issuecomment-1242721040
-        run: cargo test
+        run: cargo test -- --test-threads=1
       - name: Run grcov
         run: mkdir coverage; grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore '/*' -o ./coverage/lcov.info
       - name: Generate HTML coverage report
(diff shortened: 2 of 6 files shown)


notmandatory:
For some reason with the `default` feature set the tests aren't completing so I moved this to draft until I get that figured out.

storopoli (approved):
ACK 06f6bca34a6d301bb9738c992d7d284620585b4c

notmandatory:
To get the `default` feature tests to run without locking up I had to limit test threads to 1.

storopoli:
> To get the `default` feature tests to run without locking up I had to limit test threads to 1.

Have you tried 2?
I've managed to get away with 2 in GH Action runners in the past.

README.md
```diff
@@ -18,5 +18,8 @@ This library should compile with any combination of features with Rust 1.63.0.
 To build with the MSRV you will need to pin dependencies as follows:
 
 ```shell
+cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
```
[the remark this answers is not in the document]
notmandatory: I updated the README to specifiy you only need to pin the `jobserver` dependency to build with the MSRV. Even though this is also a dev dependency I couldn't find a way to build with MSRV without pinning it.

notmandatory:
> > To get the `default` feature tests to run without locking up I had to limit test threads to 1.
> 
> Have you tried 2? I've managed to get away with 2 in GH Action runners in the past.

Two would probably work, but it runs fast enough with 1 so I don't see a need to push it.

README.md
```diff
@@ -18,5 +18,5 @@ This library should compile with any combination of features with Rust 1.63.0.
 To build with the MSRV you will need to pin dependencies as follows:
 
 ```shell
-cargo update -p home --precise 0.5.5
+cargo update -p jobserver --precise "0.1.26"
```
oleonardolima, line 21: @notmandatory I don't think I get it 🤔 
We only need to pin the `jobserver` but we are pinning some others on the CI ?

notmandatory: We only need to pin jobserver to compile with msrv. The others are only needed to run the tests (in CI) with the msrv.

oleonardolima (approved):

evanlinjin (approved):
ACK ac2fd862ef5b457fc7e158f3f19ef64c77589136

tcharding:
For the `rust-bitcoin` upgrade I believe we need to use `electrsd` v0.27

Attempting this I'm getting a bizarre error
```bash

tcharding (approved):
ACK ac2fd862ef5b457fc7e158f3f19ef64c77589136

PR #82, opened 2024-03-21
fix hex dependency version, set to 0.2.0

notmandatory:
We need to set a version for the hex dependency, crates.io does not let us publish a package with an * version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ path = "src/lib.rs"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 bitcoin = { version = "0.31.0", features = ["serde", "std"], default-features = false }
-hex = { package = "hex-conservative", version = "*" }
+hex = { package = "hex-conservative", version = "0.2" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -19,7 +19,7 @@ path = "src/lib.rs"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 bitcoin = { version = "0.31.0", features = ["serde", "std"], default-features = false }
-hex = { package = "hex-conservative", version = "*" }
+hex = { package = "hex-conservative", version = "0.2.0" }
```
storopoli, line 22:
```suggestion
hex = { package = "hex-conservative", version = "0.2" }
```

We could use this. It has the same semantic versioning meaning but with 2 less characters.

oleonardolima (approved):
ACK b55ec14129f25a3ea33164f5812f8e6982ff2f8c _but mind storopoli's comment as well :)_

oleonardolima (approved):
ACK 811ff7e46c33c0491e7013144b974e509a4c9aa8

storopoli (approved):
ACK 811ff7e46c33c0491e7013144b974e509a4c9aa8

vladimirfomene (approved):
ACK 811ff7e46c33c0491e7013144b974e509a4c9aa8

PR #87, opened 2024-06-15
feat: add way to make raw req and get url for `BlockingClient`

evanlinjin:
Also add method for getting from `/tx/:txid` endpoint.

PR #93, opened 2024-08-21
refactor(async): refactoring async client and minor documentation improvements

oleonardolima:
### Description

It builds on top of #95. Applies minor improvements on some docstrings, and Cargo.toml standard.

The main change is adding the common HTTP methods for the `AsyncClient`, it removes duplicated code from each Esplora API request, and follows the approach done for `BlockingClient`.

It makes it easier to extract these methods into an `AsyncEsploraClient` trait (to be done in another PR, initially done here 9cbc3873c2a56258e13cb36e0e8c32039f0947a7), which the user can implement with any HTTP client of its choice. Also, makes it simpler to rebase and update the `AsyncAnonymizedClient` from #67.

### Notes to the reviewers

It has some commits from #95, as it builds on top of it and should be merged afterward. Please let me know what you think about the proposed changes and approach.

### Changelog notice

- Applies minor improvements on documentation.
- Add common `get_response` and `post_request` methods to `AsyncClient`, previously duplicated through the esplora API calls. It follows the approach done for `BlockinClient`.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -30,12 +30,14 @@ use crate::{BlockStatus, BlockSummary, Builder, Error, MerkleProof, OutputStatus
 
 #[derive(Debug, Clone)]
 pub struct AsyncClient {
+    /// The URL of the Esplora Server.
     url: String,
+    /// The inner [`reqwest::Client`] to make HTTP requests.
     client: Client,
 }
 
 impl AsyncClient {
-    /// build an async client from a builder
+    /// Build an async client from a builder
     pub fn from_builder(builder: Builder) -> Result<Self, Error> {
         let mut client_builder = Client::builder();
 
@@ -64,7 +66,7 @@ impl AsyncClient {
         Ok(Self::from_client(builder.base_url, client_builder.build()?))
     }
 
-    /// build an async client from the base url and [`Client`]
+    /// Build an async client from the base url and [`Client`]
     pub fn from_client(url: String, client: Client) -> Self {
         AsyncClient { url, client }
     }
--- a/src/async.rs
+++ b/src/async.rs
@@ -104,10 +104,7 @@ impl AsyncClient {
     async fn get_opt_response<T: Decodable>(&self, path: &str) -> Result<Option<T>, Error> {
         match self.get_response::<T>(path).await {
             Ok(res) => Ok(Some(res)),
-            Err(Error::HttpResponse { status, message }) => match status {
-                404 => Ok(None),
-                _ => Err(Error::HttpResponse { status, message }),
-            },
+            Err(Error::HttpResponse { status: 404, .. }) => Ok(None),
             Err(e) => Err(e),
         }
     }
@@ -151,20 +148,17 @@ impl AsyncClient {
     ) -> Result<Option<T>, Error> {
         match self.get_response_json(url).await {
             Ok(res) => Ok(Some(res)),
-            Err(Error::HttpResponse { status, message }) => match status {
-                404 => Ok(None),
-                _ => Err(Error::HttpResponse { status, message }),
-            },
+            Err(Error::HttpResponse { status: 404, .. }) => Ok(None),
             Err(e) => Err(e),
         }
     }
 
     /// Make an HTTP GET request to given URL, deserializing to any `T` that
-    /// implement [`bitcoin::consensus::Decodable`] from Hex, [`Vec<u8>`].
+    /// implements [`bitcoin::consensus::Decodable`].
     ///
-    /// It should be used when requesting Esplora endpoints that can be directly
-    /// deserialized to native `rust-bitcoin` types, which implements
-    /// [`bitcoin::consensus::Decodable`] from Hex, `Vec<&u8>`.
+    /// It should be used when requesting Esplora endpoints that are expected
+    /// to return a hex string decodable to native `rust-bitcoin` types which
+    /// implement [`bitcoin::consensus::Decodable`] from `&[u8]`.
     ///
     /// # Errors
     ///
@@ -194,10 +188,7 @@ impl AsyncClient {
     async fn get_opt_response_hex<T: Decodable>(&self, path: &str) -> Result<Option<T>, Error> {
         match self.get_response_hex(path).await {
             Ok(res) => Ok(Some(res)),
-            Err(Error::HttpResponse { status, message }) => match status {
-                404 => Ok(None),
-                _ => Err(Error::HttpResponse { status, message }),
-            },
+            Err(Error::HttpResponse { status: 404, .. }) => Ok(None),
             Err(e) => Err(e),
         }
     }
@@ -233,10 +224,7 @@ impl AsyncClient {
     async fn get_opt_response_text(&self, path: &str) -> Result<Option<String>, Error> {
         match self.get_response_text(path).await {
             Ok(s) => Ok(Some(s)),
-            Err(Error::HttpResponse { status, message }) => match status {
-                404 => Ok(None),
-                _ => Err(Error::HttpResponse { status, message }),
-            },
+            Err(Error::HttpResponse { status: 404, .. }) => Ok(None),
             Err(e) => Err(e),
         }
     }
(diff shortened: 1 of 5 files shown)


src/async.rs
```diff
@@ -64,33 +66,210 @@ impl AsyncClient {
...
+        let url = format!("{}{}", self.url, path);
+        let response = self.client.get(url).send().await?;
+
+        match response.status().is_success() {
+            true => Ok(deserialize::<T>(&response.bytes().await?)?),
+            false => Err(Error::HttpResponse {
+                status: response.status().as_u16(),
+                message: response.text().await?,
+            }),
+        }
```
ValuedMammal, line 95: style nit, not a blocker though

https://rust-lang.github.io/rust-clippy/master/index.html#/match_bool
```suggestion
        if response.status().is_success() {
            Ok(deserialize::<T>(&response.bytes().await?)?)
        } else {
            Err(Error::HttpResponse {
                status: response.status().as_u16(),
                message: response.text().await?,
            })
        }
```

ValuedMammal (commented):
Light review ACK

ValuedMammal:
A recent PR #87 exposed a [`get_request`](https://github.com/bitcoindevkit/rust-esplora-client/blob/d008b9b24cfd0e366728b3db52e6736447bfbb8b/src/blocking.rs#L60) method for the blocking client that returns a new `Request`, would you want to do a similar thing for the async client?

src/async.rs
oleonardolima: Thanks, I'll address it.

oleonardolima:
> A recent PR #87 exposed a [`get_request`](https://github.com/bitcoindevkit/rust-esplora-client/blob/d008b9b24cfd0e366728b3db52e6736447bfbb8b/src/blocking.rs#L60) method for the blocking client that returns a new `Request`, would you want to do a similar thing for the async client?

IIUC exposes a `get_request` because it does not have a `minreq` client as a field on `BlockingClient`, as `minreq` does not have this "strategy" as `reqwest` has :think

I think it could be helpful to get exposed in async/reqwest too, I'll check it.

src/async.rs
ValuedMammal: FWIW I find this to be quite readable, but whatever you choose is fine

```rust
    if !response.status().is_success() {
        return Err(Error::HttpResponse {
            status: response.status().as_u16(),
            message: response.text().await?,
        });
    }

    Ok(deserialize::<T>(&response.bytes().await?)?)
```

ValuedMammal:
> > A recent PR #87 exposed a [`get_request`](https://github.com/bitcoindevkit/rust-esplora-client/blob/d008b9b24cfd0e366728b3db52e6736447bfbb8b/src/blocking.rs#L60) method for the blocking client that returns a new `Request`, would you want to do a similar thing for the async client?
> 
> IIUC exposes a `get_request` because it does not have a `minreq` client as a field on `BlockingClient`, as `minreq` does not have this "strategy" as `reqwest` has :think
> 
> I think it could be helpful to get exposed in async/reqwest too, I'll check it.

Ah ok so maybe not necessary since you can already get a handle to the `reqwest::Client`?

oleonardolima:
> > > A recent PR #87 exposed a [`get_request`](https://github.com/bitcoindevkit/rust-esplora-client/blob/d008b9b24cfd0e366728b3db52e6736447bfbb8b/src/blocking.rs#L60) method for the blocking client that returns a new `Request`, would you want to do a similar thing for the async client?
> > 
> > 
> > IIUC exposes a `get_request` because it does not have a `minreq` client as a field on `BlockingClient`, as `minreq` does not have this "strategy" as `reqwest` has :think
> > I think it could be helpful to get exposed in async/reqwest too, I'll check it.
> 
> Ah ok so maybe not necessary since you can already get a handle to the `reqwest::Client`?

Yes, that's what I thought.

src/async.rs
oleonardolima: Cool! I did the update.

ValuedMammal (approved):
ACK 9cb662ae73206c3b7c2daa0c44c7585f190451f2

Overall this looks great.

In 9cb662ae73206c3b7c2daa0c44c7585f190451f2

> refactor(async)!: add common GET and POST methods

Maybe I missed it but I didn't see how this change is API breaking in any way?

src/async.rs
oleonardolima: I applied the same style to the other methods.

oleonardolima:
> ACK [9cb662a](https://github.com/bitcoindevkit/rust-esplora-client/commit/9cb662ae73206c3b7c2daa0c44c7585f190451f2)
> 
> Overall this looks great.
> 
> In [9cb662a](https://github.com/bitcoindevkit/rust-esplora-client/commit/9cb662ae73206c3b7c2daa0c44c7585f190451f2)
> 
> > refactor(async)!: add common GET and POST methods
> 
> Maybe I missed it but I didn't see how this change is API breaking in any way?

Sweet! I updated the commit message, and add a new one applying the same style on other methods.

src/async.rs
```diff
@@ -64,31 +66,210 @@ impl AsyncClient {
...
+    /// It uses [`AsyncEsploraClient::get_response`] internally.
+    ///
+    /// See [`AsyncEsploraClient::get_response`] above for full documentation.
+    async fn get_opt_response<T: Decodable>(&self, path: &str) -> Result<Option<T>, Error> {
+        match self.get_response::<T>(path).await {
+            Ok(res) => Ok(Some(res)),
+            Err(Error::HttpResponse { status, message }) => match status {
+                404 => Ok(None),
+                _ => Err(Error::HttpResponse { status, message }),
+            },
```
ValuedMammal, line 110: Would this also work?
```suggestion
            Err(Error::HttpResponse { status: 404, .. }) => Ok(None),
```

src/async.rs
```diff
@@ -64,31 +66,210 @@ impl AsyncClient {
...
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Make an HTTP GET request to given URL, deserializing to any `T` that
+    /// implement [`bitcoin::consensus::Decodable`] from Hex, [`Vec<u8>`].
+    ///
+    /// It should be used when requesting Esplora endpoints that can be directly
+    /// deserialized to native `rust-bitcoin` types, which implements
+    /// [`bitcoin::consensus::Decodable`] from Hex, `Vec<&u8>`.
```
ValuedMammal, line 167: I tried to write this in a similar way to the docs for `get_response` with the added comment that it returns a hex string.
```rs
/// Make an HTTP GET request to given URL, deserializing to any `T` that
/// implements [`bitcoin::consensus::Decodable`].
///
/// It should be used when requesting Esplora endpoints that are expected
/// to return a hex string decodable to native `rust-bitcoin` types which
/// implement [`bitcoin::consensus::Decodable`] from `&[u8]`.
```

ValuedMammal (approved):
ACK a660346f5500c4c8826e0f2d88adb8d9124d1f84

I made comments only if you plan to retouch

src/async.rs
oleonardolima: I think it's better, update the style at 1103936477a2f502658b6240ea9be8917e8f12b0

evanlinjin:
@oleonardolima since it seems we are planning to go down the #97 route, is this PR still relevant?

oleonardolima:
> @oleonardolima since it seems we are planning to go down the #97 route, is this PR still relevant?

I guess this one can still land if we'd like to cut a release with these improvements, and mainly with #98. Then, cut a new release specifically with the refactor of #97. 🤔

ValuedMammal (approved):
ACK 1103936477a2f502658b6240ea9be8917e8f12b0

notmandatory (approved):
ACK 1103936477a2f502658b6240ea9be8917e8f12b0

Overall this is a great cleanup and should be much easier to maintain.

PR #94, opened 2024-08-22
fix(ci): pin tokio to 1.38.1 on MSRV step

oleonardolima:
### Description

As previously noticed on https://github.com/bitcoindevkit/bdk/pull/1524, tokio raised their MSRV to `1.70.0` in their latest release, in short it was because they moved to `mio 1.0` which requires `1.70.0` MSRV.

Although we don't have `tokio` as a direct dependency on `rust-esplora-client` it comes as a transitive dependency from `reqwest`.

This PR adds the necessary pinning on CI and updates the instruction on `README.md` in order to properly build on MSRV.

### Changelog notice

- Adds another MSRV pinning step on CI for the `tokio 1.38.1`.

#### All Submissions:

* [x] I've signed all my commits
* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/bdk/blob/master/CONTRIBUTING.md)
* [x] I ran `cargo fmt` and `cargo clippy` before committing

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -59,6 +59,7 @@ jobs:
         cargo update -p time --precise "0.3.20"
         cargo update -p home --precise 0.5.5
         cargo update -p url --precise "2.5.0"
+        cargo update -p tokio --precise "1.38.1"
     - name: Build
       run: cargo build --features ${{ matrix.features }} --no-default-features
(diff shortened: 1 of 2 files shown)


notmandatory (approved):
ACK 16cdd384b23fdfe76c5b2934d824bdb04891d240

Thanks for fixing this here!

PR #98, opened 2024-09-07
Implement retryable calls

ValuedMammal:
Based on #93 the PR implements retryable calls for request failure due to too many requests (429) or service unavailable (503).

Inspired by #71
h/t @e1a0a0ea

### Notes

I've added the field `max_retries` to the `Builder`. `max_retries` is also added to each of `AsyncClient`, `BlockingClient`. I added the dependency `async-std` in order to have async `sleep`.

Instead of implementing a trait on the `Request` type as in #71, the approach is to add a method on the client that sends a get request to a url and returns the response after allowing for retries. I tested on the bdk `wallet_esplora_*` example crates against https://blockstream.info/testnet/api and it seemed to resolve the 429 issue.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
 reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
+async-std = { version = "1.13.0", optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
@@ -37,7 +38,7 @@ blocking-https = ["blocking", "minreq/https"]
 blocking-https-rustls = ["blocking", "minreq/https-rustls"]
 blocking-https-native = ["blocking", "minreq/https-native"]
 blocking-https-bundled = ["blocking", "minreq/https-bundled"]
-async = ["reqwest", "reqwest/socks"]
+async = ["async-std", "reqwest", "reqwest/socks"]
 async-https = ["async", "reqwest/default-tls"]
 async-https-native = ["async", "reqwest/native-tls"]
 async-https-rustls = ["async", "reqwest/rustls-tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,10 @@ path = "src/lib.rs"
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 bitcoin = { version = "0.32", features = ["serde", "std"], default-features = false }
-hex = { package = "hex-conservative", version = "0.2" }
+hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
-reqwest = { version = "0.11", optional = true, default-features = false, features = ["json"] }
+reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
--- a/src/async.rs
+++ b/src/async.rs
@@ -30,12 +30,14 @@ use crate::{BlockStatus, BlockSummary, Builder, Error, MerkleProof, OutputStatus
 
 #[derive(Debug, Clone)]
 pub struct AsyncClient {
+    /// The URL of the Esplora Server.
     url: String,
+    /// The inner [`reqwest::Client`] to make HTTP requests.
     client: Client,
 }
 
 impl AsyncClient {
-    /// build an async client from a builder
+    /// Build an async client from a builder
     pub fn from_builder(builder: Builder) -> Result<Self, Error> {
         let mut client_builder = Client::builder();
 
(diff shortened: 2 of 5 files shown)


src/blocking.rs
```diff
@@ -339,6 +350,28 @@ impl BlockingClient {
...
+    /// for retryable error codes until max retries hit.
+    pub fn get_with_retry(&self, url: &str) -> Result<Response, Error> {
+        let mut attempts = 0;
+        let mut delay = self.backoff;
+
+        loop {
+            match self.get_request(url)?.send() {
+                Ok(resp)
+                    if attempts < self.max_retries
+                        && RETRYABLE_ERROR_CODES.contains(&(resp.status_code as u16)) =>
```
oleonardolima, line 364: nit: this could be extracted to `is_status_retryable` as we do for `success` and `not_found`

oleonardolima (commented):
Concept ACK

src/async.rs
```diff
@@ -410,4 +424,26 @@ impl AsyncClient {
...
+    /// for retryable error codes until max retries hit.
+    async fn get_with_retry(&self, url: &str) -> Result<Response, Error> {
+        let mut delay = BASE_BACKOFF_MILLIS;
+        let mut attempts = 0;
+
+        loop {
+            match self.client.get(url).send().await {
+                Ok(resp)
+                    if attempts < self.max_retries
+                        && RETRYABLE_ERROR_CODES.contains(&resp.status().as_u16()) =>
```
oleonardolima, line 438: nit: You can apply the same approach here and extract it to `is_retryable_status` too.

src/async.rs
```diff
@@ -410,4 +424,26 @@ impl AsyncClient {
...
     }
+
+    /// Sends a GET request to the given `url`, retrying failed attempts
+    /// for retryable error codes until max retries hit.
+    async fn get_with_retry(&self, url: &str) -> Result<Response, Error> {
+        let mut delay = BASE_BACKOFF_MILLIS;
+        let mut attempts = 0;
+
+        loop {
+            match self.client.get(url).send().await {
```
oleonardolima, line 435: nit: You can also just propagate the error here, right?

P.S.: The same applies to the blocking function.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ name = "esplora_client"
 path = "src/lib.rs"
 
 [dependencies]
+async-std = "1.13.0"
```
oleonardolima, line 20: It should be an optional dependency, and only available when using the `async` feature.

Also, I don't think we need all the features in its default features, and only the `std` feature, is that right?

oleonardolima (commented):
It's looking good! Thanks for working on this, I just left a few comments and nits.

I wonder if we should try to add a test for the retry behavior, but unsure if it's worth it.

Cargo.toml
ValuedMammal: I tried with only `std` feature and it seems that the `task` module is only available with "default"

ValuedMammal:
- Made `async-std` an optional dep
- Small improvements to `get_with_retry`

@oleonardolima I haven't come up with a test for the retry behavior, although we can at least be confident that the methods that call `get_with_retry` still return a response

Cargo.toml
oleonardolima: Yes, looks like it's behind the default feature [here](https://docs.rs/async-std/latest/src/async_std/task/mod.rs.html#133-164) using their [`cfg_default!`](https://github.com/async-rs/async-std/blob/fa29b508e138ee0c91873039bc28cb71a4546460/src/utils.rs#L231-L238) macro.

oleonardolima (approved):
ACK 693af3989a8b8f0121e5931dc82ef6a2e4652900

nit: I'd just update the last commit message to something like: ```refactor(async,blocking): `get_with_retry` fn``` to follow the standard.

src/async.rs
```diff
@@ -455,4 +424,26 @@ impl AsyncClient {
...
+
+    /// Sends a GET request to the given `url`, retrying failed attempts
+    /// for retryable error codes until max retries hit.
+    async fn get_with_retry(&self, url: &str) -> Result<Response, Error> {
+        let mut delay = BASE_BACKOFF_MILLIS;
+        let mut attempts = 0;
+
+        loop {
+            match self.client.get(url).send().await? {
+                resp if attempts <= self.max_retries && is_status_retryable(resp.status()) => {
```
ValuedMammal, line 436: I think this was probably better using `<` rather than `<=`

oleonardolima: I don't see much problem using `<=`, although indeed using `<=` would add a few more seconds as it "adds another attempt".

oleonardolima: However, for the sake of correctness with the used constant, I'm fine we can update it `<`.

oleonardolima (approved):
ACK b96270e2993377f121fccdbf117676830f059049

notmandatory (approved):
tACK b96270e2993377f121fccdbf117676830f059049

I tested with `example_esplora` on mainnet (with gap-limit 100) and testnet. In both cases tests failed with the current version of this code and works with the fixes in this PR. 

Overall this is also a great cleanup and should be much easier to maintain.

src/blocking.rs
```diff
@@ -338,6 +346,24 @@ impl BlockingClient {
         };
         self.get_response_json(&path)
     }
+
+    /// Sends a GET request to the given `url`, retrying failed attempts
+    /// for retryable error codes until max retries hit.
+    pub fn get_with_retry(&self, url: &str) -> Result<Response, Error> {
```
ValuedMammal, line 352: Oops, I don't think it was intended for this to be `pub`.

PR #99, opened 2024-09-15
refactor(api, blocking, async)!: make library generic over any HTTP client

oleonardolima:


oleonardolima:
@evanlinjin, I've started working on this one, on top of what you've suggested if you want to have a look.

I'm still thinking of ways to have better ergonomics on its usage, and not have the multiple deserialize_... methods.

evanlinjin:
Thanks for the progress on this! I still need to have a proper look.

For now, I'm wondering whether we should just depend on the [`http`](https://docs.rs/http/latest/http/) crate for HTTP types. My argument is that `hyper`, `ureq`, `reqwest` all depend on it. `minreq` does not, but we can still convert from `http` types to `minreq` types, and I don't think we should promote it since it doesn't do H2.

oleonardolima:
> Thanks for the progress on this! I still need to have a proper look.
> 
> For now, I'm wondering whether we should just depend on the [`http`](https://docs.rs/http/latest/http/) crate for HTTP types. My argument is that `hyper`, `ureq`, `reqwest` all depend on it. `minreq` does not, but we can still convert from `http` types to `minreq` types, and I don't think we should promote it since it doesn't do H2.

Thanks! Sure, I'll update it to depend on the `HTTP` crate. It's a pretty slim dependency, and I don't see a problem with relying on it.

Agree, it's just a matter of implementing a `From` trait for the correspondent `minreq` (or any other http library) types.

PR #100, opened 2024-09-25
fix(blocking): `get_with_retry` is private method

ValuedMammal:
Clean up after #98 which accidentally exposed `get_with_retry` publicly for the blocking client.


oleonardolima (approved):
ACK bfaa9ae546e4087c1a81c653333ee9b1c949e0ce

Issue #102, opened 2024-10-03
`rust-esplora-client` shouldn't introduce an async runtime


tnull:
In #98, a dependency on the `async_std` runtime was introduced. IIUC, this was done *only* to use its `sleep` method.

I don't think `rust-esplora-client` should make choices regarding which async runtime users should use. If anything, it should be using the by now default `tokio` rather than `async_std`.

I believe I already brought this up in the discussion of the precursor #71, which however seems now gone for some reason.

An alternative to pulling in an extra dependency for this would simply be to have the user specify a `sleep` callback function that would use whatever sleep functionality they have available in their runtime.

(cc @notmandatory @ValuedMammal)

ValuedMammal:
Thanks for raising the issue @tnull

PR #103, opened 2024-10-13
Remove async std dependency 

praveenperera:
Closes: #102

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,8 +22,11 @@ bitcoin = { version = "0.32", features = ["serde", "std"], default-features = fa
 hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
-reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
-async-std = { version = "1.13.0", optional = true }
+reqwest = { version = "0.12",  features = ["json"], default-features = false, optional = true }
+
+# async runtime
+async-std = { version = "1.13", optional = true }
+tokio = { version = "1.40", features = ["time"], optional = true }
 
 [dev-dependencies]
 serde_json = "1.0"
@@ -38,7 +41,11 @@ blocking-https = ["blocking", "minreq/https"]
 blocking-https-rustls = ["blocking", "minreq/https-rustls"]
 blocking-https-native = ["blocking", "minreq/https-native"]
 blocking-https-bundled = ["blocking", "minreq/https-bundled"]
-async = ["async-std", "reqwest", "reqwest/socks"]
+
+tokio = ["dep:tokio", "async"]
+async-std = ["dep:async-std", "async"]
+
+async = ["reqwest", "reqwest/socks"]
 async-https = ["async", "reqwest/default-tls"]
 async-https-native = ["async", "reqwest/native-tls"]
 async-https-rustls = ["async", "reqwest/rustls-tls"]
--- a/src/async.rs
+++ b/src/async.rs
@@ -11,7 +11,6 @@
 
 //! Esplora by way of `reqwest` HTTP client.
 
-use async_std::task;
 use std::collections::HashMap;
 use std::str::FromStr;
 
@@ -434,7 +433,7 @@ impl AsyncClient {
         loop {
             match self.client.get(url).send().await? {
                 resp if attempts < self.max_retries && is_status_retryable(resp.status()) => {
-                    task::sleep(delay).await;
+                    crate::runtime::sleep(delay).await;
                     attempts += 1;
                     delay *= 2;
                 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,8 @@ pub mod api;
 pub mod r#async;
 #[cfg(feature = "blocking")]
 pub mod blocking;
+#[cfg(feature = "async")]
+pub mod runtime;
 
 pub use api::*;
 #[cfg(feature = "blocking")]
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -1,0 +1,8 @@
+#[cfg(feature = "tokio")]
+pub use tokio::time::sleep;
+
+#[cfg(feature = "async-std")]
+pub use async_std::task::sleep;
+
+#[cfg(not(any(feature = "tokio", feature = "async-std")))]
+compile_error!("Either 'tokio' or 'async-std' feature must be enabled");
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Install grcov
         run: if [[ ! -e ~/.cargo/bin/grcov ]]; then cargo install grcov; fi
       - name: Test
-        run: cargo test -- --test-threads=1
+        run: cargo test --features tokio -- --test-threads=1
       - name: Run grcov
         run: mkdir coverage; grcov . --binary-path ./target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore '/*' -o ./coverage/lcov.info
       - name: Generate HTML coverage report
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 Bitcoin Esplora API client library. Supports plaintext, TLS and Onion servers. Blocking or async.
 
+If using the async client, you can choose between `tokio` or `async-std` as the runtime. One of these features must be enabled.
+
 <p>
     <a href="https://crates.io/crates/esplora-client"><img alt="Crate Info" src="https://img.shields.io/crates/v/esplora-client.svg"/></a>
     <a href="https://github.com/bitcoindevkit/rust-esplora-client/blob/master/LICENSE"><img alt="MIT Licensed" src="https://img.shields.io/badge/license-MIT-blue.svg"/></a>


praveenperera:
can't seem to get this working with rust 1.63.0

src/runtime.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg(feature = "tokio")]
+pub use tokio::time::sleep;
+
+#[cfg(feature = "async-std")]
+pub use async_std::task::sleep;
+
+#[cfg(not(any(feature = "tokio", feature = "async-std")))]
```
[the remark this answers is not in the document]
praveenperera: I’m good with only supporting Tokio. 

I think WASM users would just use the blocking client and skip the async features all together?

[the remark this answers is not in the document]
praveenperera: Adding `tokio` or `async-std` would only be needed is the async features is also enabled.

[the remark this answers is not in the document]
praveenperera: Interesting, i didn't know that, i've only used WASM for any network calls just number crunching. Does their runtime have a task too? Could we use that with features and conditional compilation?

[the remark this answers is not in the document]
praveenperera: I think that makes sense, as long as we keep the logic for retires and that kinda stuff.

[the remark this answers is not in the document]
oleonardolima: I agree with tnull's comments, I think it's pretty straightforward to have and use a `Sleeper` trait, using the `tokio` (already existing one) as the default batteries included implementation, alongside allowing the user to implement their own, for example, `gloo-timers` for WASM.

praveenperera:
@tnull @oleonardolima using a trait instead now

oleonardolima:
> @tnull @oleonardolima using a trait instead now

Awesome! I'd still not bring `async-std`, and have only `tokio` as the default one, the user could use one of their choice and pass it when building the clients.

praveenperera:
Alright guys you convinced me, I never use `async-std` or see it much anymore, I removed it completely.

src/runtime.rs
```diff
@@ -0,0 +1,14 @@
+use std::time::Duration;
+
+pub trait Runtime {
+    fn sleep(duration: Duration) -> impl std::future::Future<Output = ()> + Send;
+}
```
oleonardolima, line 5: I feel that `Runtime` is too broad as we are only looking for an sleep function implementation, it's probably fine using `Sleeper` instead, here and in the `Builder` struct.

Cargo.toml
```diff
@@ -32,13 +34,15 @@ electrsd = { version = "0.28.0", features = ["legacy", "esplora_a33e97e1", "bitc
...
-default = ["blocking", "async", "async-https"]
+default = ["blocking", "async", "async-https", "tokio"]
 blocking = ["minreq", "minreq/proxy"]
 blocking-https = ["blocking", "minreq/https"]
 blocking-https-rustls = ["blocking", "minreq/https-rustls"]
 blocking-https-native = ["blocking", "minreq/https-native"]
 blocking-https-bundled = ["blocking", "minreq/https-bundled"]
-async = ["async-std", "reqwest", "reqwest/socks"]
+
+tokio = ["dep:tokio", "async"]
```
oleonardolima, line 44: I wonder if we do need to have tokio as a separate feature, maybe we can have as part of the async one, as we did for async-std, still unsure about it. WDYT ?

Cargo.toml
```diff
@@ -22,8 +22,10 @@ bitcoin = { version = "0.32", features = ["serde", "std"], default-features = fa
 hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
-reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
-async-std = { version = "1.13.0", optional = true }
+reqwest = { version = "0.12",  features = ["json"], default-features = false, optional = true }
```
oleonardolima, line 25: I think the `reqwest` version bumping should be in its own commit.

oleonardolima (commented):
I guess there are some old commits that could be dropped, or everything squashed into a single one.

I'm still unsure about the appropriate feature structure, looking forward to have other's opinions about it.

Cargo.toml
praveenperera: fixed

Cargo.toml
praveenperera: I think it makes sense, its a default feature, but if a user wanted to use a different async runtime, but still wanted to use reqwest, without tokio like in WASM, they could disable default features and add `wasm` and `async`, features.

I think it gives more flexibilty without adding annoyance since its a default feature.

oleonardolima: Yes, I was afraid of bringing some other unwanted dependencies through `tokio`, but `reqwest` already uses it with features `net` and `time`, so it should be unified, looks fine.

src/lib.rs
```diff
@@ -66,23 +66,26 @@
...
+use std::{collections::HashMap, marker::PhantomData};
 
 pub mod api;
 
 #[cfg(feature = "async")]
 pub mod r#async;
 #[cfg(feature = "blocking")]
 pub mod blocking;
+#[cfg(feature = "async")]
+pub mod sleeper;
```
oleonardolima, line 81: It's failing to build when using `--no-default-features`, in order to use only `blocking` client.

```suggestion
pub mod sleeper;
```

src/lib.rs
```diff
@@ -992,4 +1013,32 @@ mod test {
...
+    struct TestRuntime;
+
+    #[cfg(not(feature = "tokio"))]
+    impl Sleeper for TestRuntime {
+        async fn sleep(duration: Duration) {
+            tokio::time::sleep(duration).await;
+        }
+    }
+
+    #[cfg(not(feature = "tokio"))]
```
oleonardolima, line 1035:
```suggestion
    #[cfg(all(feature = "async", not(feature = "tokio")))]
```

src/lib.rs
```diff
@@ -992,4 +1013,32 @@ mod test {
...
+    fn use_builder_with_tokio_as_normal() {
+        let builder = Builder::new("https://blockstream.info/testnet/api");
+        let client = builder.build_async();
+        assert!(client.is_ok());
+    }
+
+    #[cfg(not(feature = "tokio"))]
+    struct TestRuntime;
+
+    #[cfg(not(feature = "tokio"))]
```
oleonardolima, line 1028:
```suggestion
    #[cfg(all(feature = "async", not(feature = "tokio")))]
```

src/lib.rs
```diff
@@ -992,4 +1013,32 @@ mod test {
...
+
+    #[cfg(feature = "tokio")]
+    #[test]
+    fn use_builder_with_tokio_as_normal() {
+        let builder = Builder::new("https://blockstream.info/testnet/api");
+        let client = builder.build_async();
+        assert!(client.is_ok());
+    }
+
+    #[cfg(not(feature = "tokio"))]
```
oleonardolima, line 1025:
```suggestion
    #[cfg(all(feature = "async", not(feature = "tokio")))]
```

src/lib.rs
```diff
@@ -992,4 +1013,32 @@ mod test {
         let tx_async = async_client.get_tx(&txid).await.unwrap();
         assert_eq!(tx, tx_async);
     }
+
+    #[cfg(feature = "tokio")]
```
oleonardolima, line 1017:
```suggestion
    #[cfg(all(feature = "async", feature = "tokio"))]
```

oleonardolima (commented):
It seems to be failing regarding using only `blocking` feature, I left some comments/suggestions.

Cargo.toml
```diff
@@ -23,7 +23,9 @@ hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
 reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
-async-std = { version = "1.13.0", optional = true }
+
+# default async runtime
+tokio = { version = "1.40", features = ["time"], optional = true }
```
ValuedMammal, line 28: I had to change this to version 1.38 in order to build on MSRV following the pinned deps in the README, not sure if it matters here

src/lib.rs
```diff
@@ -144,6 +148,28 @@ impl Builder {
...
+    }
+
+    /// Build a blocking client from builder
+    #[cfg(feature = "blocking")]
+    pub fn build_blocking(self) -> BlockingClient {
+        BlockingClient::from_builder(self)
+    }
+}
+
+impl<S: Sleeper> Builder<S> {
```
ValuedMammal, line 162: Should we move the methods `proxy`, `timeout`, `header` and `max_retries` to the above `impl Builder` with no additional bounds on `S`?

src/sleeper.rs
```diff
@@ -0,0 +1,14 @@
+use std::time::Duration;
+
+pub trait Sleeper {
+    fn sleep(duration: Duration) -> impl std::future::Future<Output = ()> + Send;
+}
+
+pub struct DefaultSleeper;
```
ValuedMammal, line 7: We can derive `Debug` and `Clone` for this so that `AsyncClient` can do the same. Items in this module should have doc comments also.

src/sleeper.rs
```diff
@@ -0,0 +1,14 @@
+use std::time::Duration;
+
+pub trait Sleeper {
+    fn sleep(duration: Duration) -> impl std::future::Future<Output = ()> + Send;
+}
```
ValuedMammal, line 5: I think we'd need to add `#[async_trait]` for this. As far as I know "return position `impl Trait` in traits" didn't stabilize until rust 1.75

praveenperera: @ValuedMammal yes you're right, should I just use `async_trait` or use [this build dep](https://docs.rs/version_check/latest/version_check/) to conditionally use `async_trait` on rust versions lower than 1.75 only?

ValuedMammal: I guess I vote for `async_trait`

praveenperera: sure i'll add it, and we can remove when MSRV gets bumped, when/how does that happen?

praveenperera: @ValuedMammal  done

src/lib.rs
```diff
@@ -66,23 +66,25 @@
...
 use std::time::Duration;
+use std::{collections::HashMap, marker::PhantomData};
 
 pub mod api;
 
 #[cfg(feature = "async")]
 pub mod r#async;
 #[cfg(feature = "blocking")]
 pub mod blocking;
+pub mod sleeper;
```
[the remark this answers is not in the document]
praveenperera: but we need the `Sleeper` trait to be accessible even if people aren't using `async` so they can implement it themselves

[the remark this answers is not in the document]
ValuedMammal: What feature arrangement are you referring to? I'm assuming one of "blocking" or "async" will have to be enabled, and it doesn't look like the blocking client is using the `Sleeper` trait at all but maybe I missed something. If you weren't keen on feature-gating the entire `sleeper` module, my other suggestion for getting things to compile would have been to make `async_std` a required (not optional) dependency.

src/lib.rs
```diff
@@ -133,6 +137,8 @@ pub struct Builder {
     pub headers: HashMap<String, String>,
     /// Max retries
     pub max_retries: usize,
+    /// Async runtime, trait must implement `sleep` function, default is `tokio`
+    marker: PhantomData<S>,
 }
```
praveenperera, line 142: @ValuedMammal moving Sleeper to only the `async` feature makes this not possible

src/lib.rs
```diff
@@ -111,7 +115,7 @@ pub fn convert_fee_rate(target: usize, estimates: HashMap<u16, f64>) -> Option<f
...
 
 #[derive(Debug, Clone)]
-pub struct Builder {
+pub struct Builder<S = DefaultSleeper> {
```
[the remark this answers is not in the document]
ValuedMammal: Why not move `DefaultSleeper` to lib.rs so that it's visible to all features? Granted, `DefaultSleeper` is only a `Sleeper` implementation when tokio is enabled, so it's basically meaningless in any other context. I don't really like that the generic bleeds over into the Builder struct but I'm not finding a better way to do it either.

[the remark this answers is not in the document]
praveenperera: ya that works, done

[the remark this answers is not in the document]
ValuedMammal: Troubleshooting the CI fail, I changed every instance of `#[cfg(all(feature = "blocking", feature = "async"))]` in the unit tests to use "tokio" instead of "async", but there's probably a better way to organize the test feature configs.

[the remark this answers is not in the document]
praveenperera: Fixed the test, forgot to add `async-trait` to it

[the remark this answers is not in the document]
praveenperera: I don't think this would be a good change, currently most end users will not have to deal with the traits at all. Anyone using `DefaultSleeper` will not need any code changes.

@ValuedMammal Tried an implementation with `dyn Sleeper` and was not able to keep the clone impl on the async client.

The generics make the code a bit more complex for the tradeoff of keeping things simple for the user. I think thats a worthwhile tradeoff.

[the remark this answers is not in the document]
praveenperera: As for the `async_trait` overhead, I think that mostly just boxes the future too. We can completely avoid that overhead when MSRV is increased. Or we can add some complexity now and only require `async_trait` if you're not using a version of rust that has the async_trait feature.

src/sleeper.rs
```diff
@@ -0,0 +1,14 @@
+use std::time::Duration;
+
+#[async_trait::async_trait]
+pub trait Sleeper {
```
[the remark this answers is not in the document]
praveenperera: done

src/lib.rs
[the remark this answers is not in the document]
ValuedMammal: @tnull This was the alternate approach following the idea of passing a sleep closure https://github.com/ValuedMammal/rust-esplora-client/commit/d83ea69bdb867b8f7ae313294f91751d3dd265ca. Not sure if there's a way to do this while keeping `Clone` for `AsyncClient`

ValuedMammal:
I added some suggestions in bitcoinppl/rust-esplora-client#4

src/lib.rs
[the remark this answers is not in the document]
oleonardolima: I'm not really in favor of adding yet another dependency, in this scenario the `async_trait` one, unless it's the only option.

I'd choose the closure approach over any other, as it seems way simpler and doesn't disturb the API, but, @ValuedMammal already has tried, and I also gave it a try, but I had no success with the `impl Clone` problem.

However, I don't think the `async_trait` is mandatory here, I took a look at other crates and gave it a try, and it seems to work without it at #107.

Let me know what you guys think cc @praveenperera @ValuedMammal.

praveenperera:
merged your changes, thanks @ValuedMammal

src/lib.rs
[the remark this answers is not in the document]
ValuedMammal: I would be in favor of dropping `async-trait` by adding the associated type `type Sleep: Future<Output = ()>` to the `Sleeper` trait

src/async.rs
```diff
@@ -33,16 +35,18 @@ use crate::{
...
 #[derive(Debug, Clone)]
-pub struct AsyncClient {
+pub struct AsyncClient<S = DefaultSleeper> {
     /// The URL of the Esplora Server.
     url: String,
     /// The inner [`reqwest::Client`] to make HTTP requests.
     client: Client,
     /// Number of times to retry a request
     max_retries: usize,
+
```
oleonardolima, line 45: I'd suggest adding a simple documentation string here.

src/async.rs
```diff
@@ -447,3 +452,23 @@ impl AsyncClient {
...
+/// Trait that defines the ability to sleep within an async runtime
+#[async_trait]
+pub trait Sleeper {
+    /// Wait until the specified `duration` has elapsed
+    async fn sleep(duration: Duration);
+}
+
+/// Default sleeper. Note this may only be used as a [`Sleeper`] implementation
+/// if the "tokio" feature is enabled.
```
oleonardolima, line 464: I think this line should be removed, as it's not behind the `tokio` feature.

src/async.rs
```diff
@@ -447,3 +452,23 @@ impl AsyncClient {
 fn is_status_retryable(status: reqwest::StatusCode) -> bool {
     RETRYABLE_ERROR_CODES.contains(&status.as_u16())
 }
+
+/// Trait that defines the ability to sleep within an async runtime
+#[async_trait]
+pub trait Sleeper {
+    /// Wait until the specified `duration` has elapsed
+    async fn sleep(duration: Duration);
+}
```
oleonardolima, line 461: I think the way to go is not rely on `async_trait`, and use the associated type as I mentioned in the comments, [here's a reference](https://github.com/bitcoindevkit/rust-esplora-client/pull/107/files#diff-db4e074f116ae66822e8ace789be86c3c01ffc1659520910bd702b22687d0ff6R454-R470).

src/async.rs
```diff
@@ -447,3 +452,23 @@ impl AsyncClient {
 fn is_status_retryable(status: reqwest::StatusCode) -> bool {
     RETRYABLE_ERROR_CODES.contains(&status.as_u16())
 }
+
+/// Trait that defines the ability to sleep within an async runtime
+#[async_trait]
+pub trait Sleeper {
+    /// Wait until the specified `duration` has elapsed
```
oleonardolima, line 459:
```suggestion
    /// Sleeps until the specified `duration` has elapsed.
```

src/lib.rs
```diff
@@ -178,11 +180,18 @@ impl Builder {
         BlockingClient::from_builder(self)
     }
 
-    // Build an asynchronous client from builder
-    #[cfg(feature = "async")]
+    /// Build an asynchronous client from builder
+    #[cfg(feature = "tokio")]
     pub fn build_async(self) -> Result<AsyncClient, Error> {
         AsyncClient::from_builder(self)
     }
```
oleonardolima, line 187: IMHO it should be `cfg(all(feature = "async", feature = "tokio"))]` instead, and the tokio feature should not enable the `async` one in `Cargo.toml`.

src/lib.rs
```diff
@@ -320,7 +330,9 @@ mod test {
         let blocking_client = builder.build_blocking();
 
         let builder_async = Builder::new(&format!("http://{}", esplora_url));
-        let async_client = builder_async.build_async().unwrap();
+        let async_client = builder_async
+            .build_async_with_sleeper::<DefaultSleeper>()
+            .unwrap();
```
oleonardolima, line 335: I'd use it as below instead, so we can make sure we're testing using the default `build_async` and `build_with_sleeper` in different feature matrix scenarios.

``` suggestion
        #[cfg(feature = "tokio")]
        let async_client = builder_async.build_async().unwrap();

        #[cfg(not(feature = "tokio"))]
        let async_client = builder_async
            .build_async_with_sleeper::<r#async::DefaultSleeper>()
            .unwrap();
```

oleonardolima (commented):
Thanks for working on this @praveenperera!

Sorry for the back-and-forth on the expected approach.

I guess by removing the `async-trait` dependency and better structuring the features, and it should be in a good state.

Cargo.toml
```diff
@@ -32,13 +34,15 @@ electrsd = { version = "0.28.0", features = ["legacy", "esplora_a33e97e1", "bitc
...
+default = ["blocking", "async", "async-https", "tokio"]
 blocking = ["minreq", "minreq/proxy"]
 blocking-https = ["blocking", "minreq/https"]
 blocking-https-rustls = ["blocking", "minreq/https-rustls"]
 blocking-https-native = ["blocking", "minreq/https-native"]
 blocking-https-bundled = ["blocking", "minreq/https-bundled"]
-async = ["async-std", "reqwest", "reqwest/socks"]
+
+tokio = ["dep:tokio"]
+async = ["reqwest", "reqwest/socks"]
```
oleonardolima, line 45: I guess you can probably use something like this, where it'll activate the tokio/time feature if tokio feature has been used.

```suggestion
async = ["reqwest", "reqwest/socks", "tokio?/time"]
```

oleonardolima (commented):
utACK

It should be ready to go, just squash the commits into a single one as just the latest ones are applicable.

oleonardolima:
Im not sure, but I would use refactor(retry)!: ..., as I consider it a breaking change on how the feature works.

ValuedMammal:
I think we will need #110 to get this to run in CI.

ValuedMammal:
@praveenperera Can you rebase?

ValuedMammal (approved):
ACK 9963f273db99629d38eea2387278ca2f64ed0456

ValuedMammal:
Code review ACK

It could use some documentation improvements, but that can be done in a follow up PR. I tested upgrading `bdk_esplora` to use this branch and it seems to work. For the record I did not try using the async client with a runtime other than tokio, but I'm interested in confirming if this change will be suitable to use for example in a WASM environment.

Cargo.toml
```diff
@@ -23,7 +23,9 @@ hex = { version = "0.2", package = "hex-conservative" }
 log = "^0.4"
 minreq = { version = "2.11.0", features = ["json-using-serde"], optional = true }
 reqwest = { version = "0.11",  features = ["json"], default-features = false, optional = true }
-async-std = { version = "1.13.0", optional = true }
+
+# default async runtime
+tokio = { version = "1.38", features = ["time"], optional = true }
```
MaxFangX, line 28: Would you mind relaxing the Tokio requirement to `1.36` or lower? `tokio::time::sleep` has been around since Tokio at least `1.0.0`.

(Asking because we happen to use 1.36 and Tokio is a pain to update because we maintain a Tokio patchset for SGX)

MaxFangX:
I can confirm that this patch compiles and works in our downstream code after reducing the tokio requirement from 1.38->1.36.

src/lib.rs
```diff
@@ -71,8 +71,10 @@ use std::fmt;
 use std::num::TryFromIntError;
 use std::time::Duration;
 
-pub mod api;
+#[cfg(feature = "async")]
+use r#async::Sleeper;
```
oleonardolima, line 75: nit: we wouldn't need this other import if we use the full `r#async::Sleeper` below instead. Could be addressed in a future PR.

oleonardolima (commented):
utACK 9963f273db99629d38eea2387278ca2f64ed0456

Cargo.toml
oleonardolima: I don't see much problem with this suggestion, we could use as in BDK: `tokio = { version = "1", features = ["time"], optional = true }`. Any thoughts @ValuedMammal ?

src/lib.rs
oleonardolima, line 75: @praveenperera I was running some standalone tests with custom runtimes: async-std and gloo-timers and noticed this. It should be `pub use` instead, right?

praveenperera: @oleonardolima yup, made the change

Cargo.toml
praveenperera: @oleonardolima @MaxFangX changed it to "1"

ValuedMammal (approved):
reACK  27c41258cfa10b5a88598364a6605b8502f79a8e

PR #112, opened 2024-11-22
chore: bump version to 0.11.0

notmandatory:

PR #113, opened 2024-12-03
Disallow empty blocks response

rustaceanrob:
This is my first attempt at resolving the empty blocks response at the client level. I didn't find that any of the current error variants accurately described the problem, but could use some outside opinions

@oleonardolima @notmandatory

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,6 +225,8 @@ pub enum Error {
     InvalidHttpHeaderName(String),
     /// Invalid HTTP Header value specified
     InvalidHttpHeaderValue(String),
+    /// The server sent an invalid response
+    InvalidResponse,
 }
 
 impl fmt::Display for Error {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -225,6 +225,8 @@ pub enum Error {
     InvalidHttpHeaderName(String),
     /// Invalid HTTP Header value specified
     InvalidHttpHeaderValue(String),
+    /// Response expected to be non-empty
+    EmptyResponse,
```
ValuedMammal, line 229: Maybe this?
```suggestion
    /// The server sent an invalid response
    InvalidResponse,
```

rustaceanrob: Sure that is more generic

rustaceanrob:
Ok updated and should pass CI

ValuedMammal (approved):
ACK 340cd1cb97c6d9e3b08306b69eb50bc16a4cab19

@notmandatory Do you want to include this in #112?

oleonardolima (approved):
utACK 340cd1cb97c6d9e3b08306b69eb50bc16a4cab19

PR #114, opened 2025-01-22
feat(client): add async and blocking clients to submit txs package

acidbunny21:


src/api.rs
```diff
@@ -123,6 +125,34 @@ pub struct AddressTxsSummary {
     pub tx_count: u32,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct SubmitPackageResult {
+	pub package_msg: String,
```
ErikDeSmedt, line 130: Tabs and whitespaces are off

src/api.rs
```diff
@@ -123,6 +125,34 @@ pub struct AddressTxsSummary {
...
+    #[serde(rename = "other-wtxid")]
+	pub other_wtxid: Option<String>,
+	pub vsize: Option<u32>,
+	pub fees: Option<MempoolFeesSubmitPackage>,
+	pub error: Option<String>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct MempoolFeesSubmitPackage {
+	pub base: f64,
```
ErikDeSmedt, line 149: What is `base` here?
Can you leave a comment

src/async.rs
```diff
@@ -363,6 +362,50 @@ impl<S: Sleeper> AsyncClient<S> {
...
+		);
+
+		if let Some(maxfeerate) = maxfeerate {
+			request = request.query(&[("maxfeerate", maxfeerate.to_string())])
+		}
+
+		if let Some(maxburnamount) = maxburnamount {
+			request = request.query(&[("maxburnamount", maxburnamount.to_string())])
+		}
+
```
ErikDeSmedt, line 397: To my understanding the `maxreerate` and `maxburnamount` are a protection measure. It should protect the user against accidentally wasting all their funds.

I'd prefer to do make the checks locally before sending the request. There is are some varieties in `electrs` implementation and I don't want users to get surprised

src/blocking.rs
```diff
@@ -295,6 +300,53 @@ impl BlockingClient {
...
+			.to_vec()
+		);
+
+		if let Some(maxfeerate) = maxfeerate {
+			request = request.with_param("maxfeerate", maxfeerate.to_string())
+		}
+
+		if let Some(maxburnamount) = maxburnamount {
+			request = request.with_param("maxburnamount", maxburnamount.to_string())
+		}
```
ErikDeSmedt, line 335: Same comment as for the `sync` example.

Please do verify the amounts client-side before sending the request

ErikDeSmedt (changes requested):

stevenroose (commented):
looks good, let just some nits. can confirm that the calls match the endpoint added in my pr on esplora (https://github.com/Blockstream/electrs/pull/119)

src/async.rs
```diff
@@ -363,6 +363,56 @@ impl<S: Sleeper> AsyncClient<S> {
...
 
+    /// Broadcast a package of [`Transaction`] to Esplora
+    ///
+    /// if `maxfeerate` is provided, any transaction whose
+    /// fee is higher will be rejected
+    ///
+    /// if  `maxburnamount` is provided, any transaction
+    /// with higher provably unspendable outputs amount
+    /// will be rejected
+    pub async fn broadcast_package(
```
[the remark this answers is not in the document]
acidbunny21: updated

src/api.rs
```diff
@@ -123,6 +125,34 @@ pub struct AddressTxsSummary {
...
+    pub package_msg: String,
+    #[serde(rename = "tx-results")]
+    pub tx_results: HashMap<String, TxResult>,
+    #[serde(rename = "replaced-transactions")]
+    pub replaced_transactions: Option<Vec<String>>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct TxResult {
+    pub txid: String,
```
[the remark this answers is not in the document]
acidbunny21: updated. I also added the same comments on the fields as the ones in rust-bitcoin

Cargo.toml
```diff
@@ -18,6 +18,7 @@ path = "src/lib.rs"
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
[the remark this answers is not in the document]
acidbunny21: done

oleonardolima (commented):
It could use a test with mempool/electrs implementation, while it's not available in blockstream/electrs and electrsd, as I've looked on the other PR and looks like it's following the same API.

src/api.rs
```diff
@@ -123,6 +124,53 @@ pub struct AddressTxsSummary {
     pub tx_count: u32,
 }
 
+#[derive(Deserialize, Debug)]
+pub struct SubmitPackageResult {
+    /// The transaction package result message. "success" indicates all transactions were accepted
+    /// into or are already in the mempool.
+    pub package_msg: String,
```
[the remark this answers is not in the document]
acidbunny21: i agree that would be best, but as I can see [here](https://github.com/rust-bitcoin/corepc/blob/b62e8c84e3271e57b10849d27191c0357711660b/types/src/model/raw_transactions.rs#L21), there is no enum defined for that field. I'd be happy to update that part as soon as it is upgraded to one there

acidbunny21:
> It could use a test with mempool/electrs implementation, while it's not available in blockstream/electrs and electrsd, as I've looked on the other PR and looks like it's following the same API.

I'm okay with adding a test, but then we need to bump electrsd to 0.30 which introduce some breaking changes on bitcoind. Might be good to first bump it in a separated PR

src/async.rs
```diff
@@ -363,6 +363,56 @@ impl<S: Sleeper> AsyncClient<S> {
...
+        &self,
+        transactions: &[Transaction],
+        maxfeerate: Option<f64>,
+        maxburnamount: Option<f64>,
+    ) -> Result<SubmitPackageResult, Error> {
+        let url = format!("{}/txs/package", self.url);
+
+        let serialized_txs = transactions
+            .iter()
+            .map(|tx| serialize(&tx).to_lower_hex_string())
```
oleonardolima, line 384: You can use `serialize_hex` directly.

src/async.rs
```diff
@@ -363,6 +363,56 @@ impl<S: Sleeper> AsyncClient<S> {
...
+            .collect::<Vec<_>>();
+
+        let mut request = self.client.post(url).body(
+            serde_json::to_string(&serialized_txs)
+                .unwrap()
+                .as_bytes()
+                .to_vec(),
```
oleonardolima, line 391: Can't we just send the hex as we do for broadcast tx, it looks like the parsing in `blockstream/electrs` PR is the same as broadcast tx one.

src/async.rs
```diff
@@ -363,6 +363,56 @@ impl<S: Sleeper> AsyncClient<S> {
...
+            return Err(Error::HttpResponse {
+                status: response.status().as_u16(),
+                message: response.text().await?,
+            });
+        }
+
+        response
+            .json::<SubmitPackageResult>()
+            .await
+            .map_err(Error::Reqwest)
```
oleonardolima, line 413: I'm quite divided if we should repurpose `post_request_hex` to handle this too, instead of having the logic here, or maybe have another `post_request_bytes` if we stick to it.

src/async.rs
acidbunny21: mmmh i don't think so, the parsing in `blockstream/electrs` PR expects a json array so we cannot simply send a single hex

src/async.rs
acidbunny21: mmmh i couldn't find a good way to factorize them, mainly because submitpackage request takes some query params

src/async.rs
acidbunny21: done

src/async.rs
oleonardolima: Great, thanks! I'll do another round of review.

acidbunny21:
@oleonardolima it looks like the pipeline failures are caused by something unrelated with the PR, could you help please? :)

oleonardolima:
> @oleonardolima it looks like the pipeline failures are caused by something unrelated with the PR, could you help please? :)

It's caused due to recent `native-tls` release which bumped their MSRV to `1.80.0`, it should be fine after #116 lands and a rebase should fix it.

stevenroose:
I think we uncovered a bug in here:
```
Reqwest(reqwest::Error { kind: Decode, source: Error("invalid type: floating point `0.0`, expected u64", line: 1, column: 230) })
```

I suspect it's in this field: `pub base: Amount,`. You should never use `Amount`'s default deserialization. It has several options, but bitcoin does amounts as floating points:
```rust
#[serde(with = "amount::serde::as_btc")]
pub base: Amount,
```
should work

oleonardolima:
@acidbunny21 Could you rebase and sign the commits on this one?

luisschwab (changes requested):
I didn't do any testing yet, but there's documentation missing on structs and missing trailing dots. Docs should follow the standard set on #147.

src/api.rs
```diff
@@ -246,6 +247,58 @@ pub struct MempoolRecentTx {
     pub value: u64,
 }
 
+#[derive(Deserialize, Debug)]
```
luisschwab: Struct doccomment missing here

src/blocking.rs
```diff
@@ -270,21 +289,63 @@ impl BlockingClient {
...
-        if let Some(timeout) = &self.timeout {
-            request = request.with_timeout(*timeout);
+    /// Broadcast a package of [`Transaction`] to Esplora
+    ///
+    /// if `maxfeerate` is provided, any transaction whose
+    /// fee is higher will be rejected
+    ///
+    /// if  `maxburnamount` is provided, any transaction
+    /// with higher provably unspendable outputs amount
+    /// will be rejected
```
luisschwab:
```suggestion
    /// Broadcast a package of [`Transaction`]s to Esplora.
    ///
    /// If `maxfeerate` is provided, any transaction whose
    /// fee is higher will be rejected.
    ///
    /// If  `maxburnamount` is provided, any transaction
    /// with higher provably unspendable outputs amount
    /// will be rejected.
```

oleonardolima (commented):
tACK 7992b34b0b11cc60b1fd7678b9e4656810fcf9a0

I've tested it with mempool.space API for testnet4, and it worked just fine.

src/async.rs
```diff
@@ -365,8 +372,49 @@ impl<S: Sleeper> AsyncClient<S> {
...
 
     /// Broadcast a [`Transaction`] to Esplora
-    pub async fn broadcast(&self, transaction: &Transaction) -> Result<(), Error> {
-        self.post_request_hex("/tx", transaction).await
+    pub async fn broadcast(&self, transaction: &Transaction) -> Result<Txid, Error> {
+        let body = serialize::<Transaction>(transaction).to_lower_hex_string();
+        let response = self.post_request_bytes("/tx", body, None).await?;
+        let txid = Txid::from_str(&response.text().await?).map_err(|_| Error::InvalidResponse)?;
+        Ok(txid)
+    }
```
oleonardolima: I just noticed that landing this PR has a breaking change, due to this change here and in the blocking client.

@tnull Do you need the submit_package API as a non-breaking release?

src/blocking.rs
```diff
@@ -272,20 +291,58 @@ impl BlockingClient {
...
+            Err(e) => Err(Error::Minreq(e)),
         }
+    }
 
-        if let Some(timeout) = &self.timeout {
-            request = request.with_timeout(*timeout);
+    /// Broadcast a package of [`Transaction`] to Esplora
+    ///
+    /// If `maxfeerate` is provided, any transaction whose
+    /// fee is higher will be rejected
```
luisschwab:
```suggestion
    /// fee is higher will be rejected.
```

src/async.rs
```diff
@@ -366,7 +373,49 @@ impl<S: Sleeper> AsyncClient<S> {
...
     pub async fn broadcast(&self, transaction: &Transaction) -> Result<(), Error> {
-        self.post_request_hex("/tx", transaction).await
+        let body = serialize::<Transaction>(transaction).to_lower_hex_string();
+        match self.post_request_bytes("/tx", body, None).await {
+            Ok(_resp) => Ok(()),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Broadcast a package of [`Transaction`] to Esplora
```
luisschwab:
```suggestion
    /// Broadcast a package of [`Transaction`]s to Esplora.
```

src/async.rs
```diff
@@ -366,7 +373,49 @@ impl<S: Sleeper> AsyncClient<S> {
 
     /// Broadcast a [`Transaction`] to Esplora
```
luisschwab:
```suggestion
    /// Broadcast a [`Transaction`] to Esplora.
```

src/async.rs
```diff
@@ -366,7 +373,49 @@ impl<S: Sleeper> AsyncClient<S> {
...
+        match self.post_request_bytes("/tx", body, None).await {
+            Ok(_resp) => Ok(()),
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Broadcast a package of [`Transaction`] to Esplora
+    ///
+    /// If `maxfeerate` is provided, any transaction whose
+    /// fee is higher will be rejected
```
luisschwab:
```suggestion
    /// fee is higher will be rejected.
```

luisschwab (changes requested):

ValuedMammal:
ACK 970ea94faef9506a07d8cb5855c3d3764d044295

luisschwab (approved):
ACK f7fbaa4c8df6eb56643f3859a9a1da2dd2aef200

PR #115, opened 2025-02-14
fix(api): allow parsing empty TXIDs

torkelrogstad:
In some cases, `mempool/electrs` returns an empty string for TXIDs in the input list when querying for address transactions. 

TXID: `c138e61852baf252d81056f6a779d84d65efc830c202215d7b904d5b1e546039`
Address: `tb1qp5dt5a96udwds8kmjz0vd2yy0q3ltn6cx56dy7`

Observe that the response from `curl https://mempool.drivechain.live/api/address/tb1qp5dt5a96udwds8kmjz0vd2yy0q3ltn6cx56dy7/txs` (including the relevant part below) includes an input with `txid` set to the empty string. I'm not sure what precisely causes this. However, it strikes me as OK in being a bit more lenient with what we accept as input without crashing and burning.

```json
{
    "txid": "c138e61852baf252d81056f6a779d84d65efc830c202215d7b904d5b1e546039",
    "version": 2,
    "locktime": 0,
    "size": 300,
    "weight": 1092,
    "fee": 0,
    "vin": [
      {
        "is_coinbase": true,
        "prevout": null,
        "scriptsig": "02f11900",
        "scriptsig_asm": "",
        "sequence": 4294967295,
        "txid": "",
        "vout": 0,
        "witness": [
          "0000000000000000000000000000000000000000000000000000000000000000"
        ],
        "inner_redeemscript_asm": "",
        "inner_witnessscript_asm": ""
      }
    ],
    "vout": [
      {
        "value": 0,
        "scriptpubkey": "6a06d77d177601ff",
        "scriptpubkey_address": "",
        "scriptpubkey_asm": "OP_RETURN OP_PUSHBYTES_6 d77d177601ff",
        "scriptpubkey_type": "op_return"
      },
      {
        "value": 5000000000,
        "scriptpubkey": "00140d1aba74bae35cd81edb909ec6a8847823f5cf58",
        "scriptpubkey_address": "tb1qp5dt5a96udwds8kmjz0vd2yy0q3ltn6cx56dy7",
        "scriptpubkey_asm": "OP_0 OP_PUSHBYTES_20 0d1aba74bae35cd81edb909ec6a8847823f5cf58",
        "scriptpubkey_type": "v0_p2wpkh"
      },
      {
        "value": 0,
        "scriptpubkey": "6a24aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf94c70ecc7daa2000247304402203cf68802e97050cd98815ac57cf6a09585a2833f5d66f7a696e2b9fa04c8628902205079d5d3327d5e1ee14974e66fcb72be24ef1872ee99cf922826fd740ff3500501210360a9e39c533bd8222088de8cb20aabccafac41627a76291bc6166b7988b11481",
        "scriptpubkey_address": "",
        "scriptpubkey_asm": "OP_RETURN OP_PUSHBYTES_36 aa21a9ede2f61c3f71d1defd3fa999dfa36953755c690689799962b48bebd836974e8cf9 OP_PUSHDATA1 ecc7daa2000247304402203cf68802e97050cd98815ac57cf6a09585a2833f5d66f7a696e2b9fa04c8628902205079d5d3327d5e1ee14974e66fcb72be24ef1872ee99cf922826fd740ff3500501210360a9e39c533bd8222088de8cb20aabccafac41627a76291bc6166b7988b11481",
        "scriptpubkey_type": "op_return"
      }
    ],
    "status": {
      "confirmed": true,
      "block_height": 6641,
      "block_hash": "0000025a4222d1c99aab47a04c4e9388fb2643023d83fe4e5beb99260f5ae6f9",
      "block_time": 1737030431
    }
  }
```

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -10,6 +10,7 @@ pub use bitcoin::{
 };
 
 use serde::Deserialize;
+use serde::Deserializer;
 
 #[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct PrevOut {
@@ -19,6 +20,7 @@ pub struct PrevOut {
 
 #[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct Vin {
+    #[serde(deserialize_with = "deserialize_txid")]
     pub txid: Txid,
     pub vout: u32,
     // None if coinbase
@@ -68,6 +70,7 @@ pub struct BlockStatus {
 
 #[derive(Deserialize, Clone, Debug, PartialEq, Eq)]
 pub struct Tx {
+    #[serde(deserialize_with = "deserialize_txid")]
     pub txid: Txid,
     pub version: i32,
     pub locktime: u32,
@@ -197,4 +200,18 @@ where
         .map(|hex_str| Vec::<u8>::from_hex(&hex_str))
         .collect::<Result<Vec<Vec<u8>>, _>>()
         .map_err(serde::de::Error::custom)
+}
+
+fn deserialize_txid<'de, D>(deserializer: D) -> Result<Txid, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    use bitcoin::hashes::Hash;
+    use std::str::FromStr;
+
+    let s = String::deserialize(deserializer)?;
+    if s.is_empty() {
+        return Ok(Txid::all_zeros());
+    }
+    Txid::from_str(&s).map_err(serde::de::Error::custom)
 }


oleonardolima:
Thanks, that's a good catch. I didn't see this behavior before.

Are you running a custom mempool/esplora API server ? I'd like to research a bit if it's an isolated behavior or if it's also happening on mainnet/testnet.

torkelrogstad:
I'm running an instance of https://github.com/mempool/electrs on a custom signet

notmandatory (approved):
utACK c6abeaefd6ab2b6c15c15e8cc1e258777088ed39

It looks like this happens when requesting a coinbase tx which understandably don't have any inputs. We probably haven't seen this before because our user's wallets don't generally include coinbase TX. But this is something we should handle gracefully.

ValuedMammal (changes requested):
This approach looks wrong.

I want to investigate if this is an isolated instance or if mainnet servers are similarly affected.

notmandatory:
As @ValuedMammal pointed out this looks like an issue with the `https://mempool.drivechain.live/api/address` implementation. You can see in the response to `curl https://mempool.space/api/tx/4256321f15100fedb2380c856799e26e95b627b997e4bbe8cb69bac33cf45782 | jq` that mempool.space already adds all zeros for the coinbase `vin.txid`. 

@torkelrogstad can we close this PR?

PR #118, opened 2025-03-11
chore(deps): update `reqwest` to `0.12`

phlip9:
Updating to LDK 0.1, I noticed `esplora-client` is the last dep we have using `reqwest-v0.11`, which forces us to hold on to a ton of duplicate, old deps in our tree (http 0.2, hyper 0.14, rustls 0.21, ...). Updating `reqwest` to `0.12` removes all these for us.

I've also updated the 1.63.0 MSRV instructions to pin `reqwest-v0.12.9`, the last version with that MSRV. Later `reqwest` versions bump the MSRV to 1.64.0.

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -54,16 +54,21 @@ jobs:
     - name: Pin dependencies for MSRV
       if: matrix.rust.version == '1.63.0'
       run: |
+        cargo update -p reqwest --precise "0.12.4"
+        cargo update -p minreq --precise "2.13.2"
         cargo update -p zstd-sys --precise "2.0.8+zstd.1.5.5"
         cargo update -p time --precise "0.3.20"
-        cargo update -p home --precise 0.5.5
+        cargo update -p home --precise "0.5.5"
         cargo update -p url --precise "2.5.0"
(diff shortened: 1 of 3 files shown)


phlip9:
Update: `tokio-util` and `indexmap` are no longer in the dependency tree, so remove those from the MSRV

phlip9:
Finally buckled down and got everything building w/ 1.63 locally.

- Needed to pin `reqwest` to `v0.12.4` to avoid all the new h3 deps.
- The new `minreq` version that released yesterday also silently started break on 1.63, so pin that to `v2.13.2`.
- New `once_cell` release also breaks MSRV, so pin to `v1.20.3`.
- Pin `ring` to `v0.17.12`.

ValuedMammal:
@phlip9 Thanks for looking into it, I'll try to get things working on my end as well

phlip9:
Ahh looks like some dev-dependencies also need to be pinned...

phlip9:
Pinned `base64-ct` to `v1.6.0` and `bzip2-sys` to `v0.1.12+1.0.8`, but only in CI since these are transitive dev-dependencies.

ValuedMammal (approved):
ACK d84d8d04897d0ce1311d90b2f017df844a06a615

PR #120, opened 2025-03-18
wip,deps: bump `electrsd` to 0.32

ValuedMammal:


ValuedMammal:
For unit tests

- using `deserialize_hex` instead of FromHex
- add back `AddressType::Legacy`

src/lib.rs
```diff
@@ -591,11 +553,21 @@ mod test {
...
+            tx_res
+                .fee
+                .map(|fee| Amount::from_btc(fee.abs()).unwrap())
+                .unwrap()
+        );
         assert!(tx_info.status.confirmed);
-        assert_eq!(tx_info.status.block_height, tx_res.info.blockheight);
-        assert_eq!(tx_info.status.block_hash, tx_res.info.blockhash);
-        assert_eq!(tx_info.status.block_time, tx_res.info.blocktime);
+        // TODO(corepc): No .block_height field on GetTransaction ?
```
oleonardolima, line 564: Oh, it's unfortunate, might take the opportunity and add it too as I need to finish my work on adding the other mining RPCs required for the BDK upgrade.

oleonardolima (commented):
utACK 0e02d793beb91684fe4bee77a678703967dc85b7

src/lib.rs
```diff
@@ -591,11 +553,21 @@ mod test {
...
+        assert_eq!(tx_info.status.block_height, height);
+        assert_eq!(
+            tx_info.status.block_hash.map(|hash| hash.to_string()),
+            tx_res.block_hash
+        );
+        assert!(tx_info.status.block_time >= Some(tx_res.time as u64));
```
oleonardolima, line 570: ~~Any reason to assert with `>=` and against `tx.time` instead ?~~

I've tried to use the available `block_time`, but didn't understand why it now keeps being off-by-one 🤔.

oleonardolima (approved):
tACK 0e02d793beb91684fe4bee77a678703967dc85b7

It looks pretty straight-forward, with the two caveats of `block_height` and `block_time` fields.

src/lib.rs
ValuedMammal: I was confused by that too, might need investigating.

oleonardolima: I have some suggested changes here: https://github.com/oleonardolima/rust-esplora-client/commit/e732a5b465364424da7078f7ab1c12f030b96a58, that reduce a little the usage of parsing, relying on corepc methods instead.

oleonardolima: FWIW using the `block_time` now seems to be working just fine, I guess the off-by-one error I was getting was related to the `block_height`, which fetching by `block_hash` solves the issue.

oleonardolima (approved):
tACK e732a5b465364424da7078f7ab1c12f030b96a58

ValuedMammal:
ACK e732a5b465364424da7078f7ab1c12f030b96a58

src/lib.rs
ValuedMammal: Since we can verify the block height by doing separate API call, I removed the comment.

ValuedMammal:
Attempted to update `electrsd` to 0.34.0 but got blocked by CI. https://github.com/bitcoindevkit/rust-esplora-client/actions/runs/15977003057/job/45062009906

oleonardolima (approved):
ACK cb6fdd2fd4ad2d94ed2f596656d95b592ddab7e9

PR #124, opened 2025-03-22
Replaced reqwest with async-minreq for async HTTP requests #121

### LINKED ISSUE #121: Investigate using `async-minreq`
**Describe the enhancement**  

Investigate and prototype using [async-minreq](https://github.com/BEULAHEVANJALIN/async-minreq) instead of `reqwest`.

**Use case**  

We've had problems in the past with the `reqwest` MSRV being higher than the `bdk` MSRV. Using a simpler crate like `async-minreq` should solve those.  

Should also investigate if `async-minreq` can be used with a different socket transport so it can be integrated with TOR via `arti`.

**Additional context**

We already using `minreq` for the `blocking` feature.

rajanarahul93:
## Summary  
Replaced `reqwest` with `async-minreq` in `rust-esplora-client` to resolve MSRV conflicts with `bdk`.  

## Changes  
- Switched async HTTP requests to `async-minreq`.  
- Updated tests and verified compatibility with Rust   
- Ensured proxy options work with TOR.  

## Related Issue  
Closes #121

### DIFF
--- a/.github/workflows/cont_integration.yml
+++ b/.github/workflows/cont_integration.yml
@@ -54,8 +54,7 @@ jobs:
     - name: Pin dependencies for MSRV
       if: matrix.rust.version == '1.63.0'
       run: |
-        cargo update -p reqwest --precise "0.12.4"
-        cargo update -p minreq --precise "2.13.2"
+        cargo update -p async-minreq --precise "2.0.0"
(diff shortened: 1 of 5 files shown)


rajanarahul93:
Looking forward to any feedback or suggestions.
@notmandatory

PR #127, opened 2025-05-18
implemented /mempool endpoint, can now access mempool info such as the fee historgram

shanmukanaks:


notmandatory:
Which versions of the Esplora server support this?

PR #134, opened 2025-08-07
feat: add `get_address_utxos` method

### LINKED ISSUE #133: feat: `/address/:address/utxo`
Implementing this would be very useful, although `rust-bitcoin` does not implement types for a UTXO following the Esplora API Spec, which is a blocker. I opened [rust-bitcoin#4761](https://github.com/rust-bitcoin/rust-bitcoin/issues/4761) for this.

luisschwab:
Closes #133.

This PR adds the `get_address_utxos` method, which hits the `/address/{addr}/utxo` endpoint, returning a `Vec<Utxo>`.

# Changelog
- Create `UtxoStatus`.
- Create `Utxo`.
- Create `get_address_utxos` for both blocking and async clients.
- Add a functional test for it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1037,4 +1037,32 @@ mod test {
         assert_eq!(address_txs_blocking, address_txs_async);
         assert_eq!(address_txs_async[0].txid, txid);
     }
+
+    #[cfg(all(feature = "blocking", feature = "async"))]
+    #[tokio::test]
+    async fn test_get_address_utxos() {
+        let (blocking_client, async_client) = setup_clients().await;
+
+        let address = BITCOIND
+            .client
+            .new_address_with_type(AddressType::Legacy)
+            .unwrap();
+
+        let _txid = BITCOIND
+            .client
+            .send_to_address(&address, Amount::from_sat(21000))
+            .unwrap()
+            .txid()
+            .unwrap();
+
+        let _miner = MINER.lock().await;
+        generate_blocks_and_wait(1);
+
+        let address_utxos_blocking = blocking_client.get_address_utxos(&address).unwrap();
+        let address_utxos_async = async_client.get_address_utxos(&address).await.unwrap();
+
+        assert_ne!(address_utxos_blocking.len(), 0);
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1037,4 +1037,32 @@ mod test {
...
+
+        let _miner = MINER.lock().await;
+        generate_blocks_and_wait(1);
+
+        let address_utxos_blocking = blocking_client.get_address_utxos(&address).unwrap();
+        let address_utxos_async = async_client.get_address_utxos(&address).await.unwrap();
+
+        assert_ne!(address_utxos_blocking.len(), 0);
+        assert_ne!(address_utxos_async.len(), 0);
+        assert_eq!(address_utxos_blocking, address_utxos_async);
```
oleonardolima, line 1066: You could also add assertions for each field.

oleonardolima (commented):
tACK 86d3d4473eef0f401f9d5cf840ba15bd1d1090c7

I've tested it with blocking client, and it worked just fine on both happy path, too many histories error and when it's unconfirmed.

src/api.rs
```diff
@@ -123,6 +123,33 @@ pub struct AddressTxsSummary {
     pub tx_count: u32,
 }
 
+/// Information about an UTXO's status: confirmation status,
+/// confirmation height, confirmation block hash and confirmation block time.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Deserialize)]
+pub struct UtxoStatus {
```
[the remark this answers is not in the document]
luisschwab: Indeed, structurally they are the same. But I preferred to create another struct with a name that matches what it models. We could change the name of `TxStatus` to something generic, but I don't think that's very elegant.

src/lib.rs
oleonardolima: I was primarily concerned about how the esplora server was returning the type (e.g, float in BTC, in sats, ...) for the value field, but my test indicates it's working as expected. Specific field assertions can also be added in a follow-up PR, if needed.

PR #140, opened 2025-11-18
feat: add `justfile`

luisschwab:
This PR adds a minimal `justfile` stolen from `bdk_wallet`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,0 +1,32 @@
+alias b := build
+alias c := check
+alias f := fmt
+alias t := test
+alias p := pre-push
+
+_default:
+   @just --list
+
+# Build the project
+build:
+   cargo build
+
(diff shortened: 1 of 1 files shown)


oleonardolima (commented):
utACK cd6238a453fe31e3d769d36171cd224e969045ee

justfile
```diff
@@ -0,0 +1,32 @@
...
+       echo "\n⚠️  Unsigned commit: BDK requires that commits be signed." || \
+       true
+
+# Format all code
+fmt:
+   cargo +nightly fmt
+
+# Run all tests on the workspace with all features
+test:
+   cargo test --all-features
```
oleonardolima: At least on macOS it's failing intermitently, and I usually use `test -- --test-threads=2`, but in CI we use `test -- --test-threads=1`.

PR #145, opened 2025-12-08
Add Async OHTTP client

arminsabouri:
[OHTTP](https://datatracker.ietf.org/doc/rfc9458/) lets a client send encrypted requests through a relay so the server can’t see who sent them and the relay can’t see what they contain. The following commit adds optional configurations to enable clients to proxy their requests through an OHTTP relay and gateway.

OHTTP functionality is feature flagged off behind `async-ohttp`. If a client provided OHTTP config it will attempt to use the relay instead of the target resource directly.

---- 
This work was [first started](https://github.com/alexlwn123/rust-esplora-client/pull/1) by @alexlwn123
Tested against a blockstream's [electrs](https://github.com/Blockstream/electrs/tree/new-index) and running an [OHTTP relay](https://github.com/payjoin/ohttp-relay) and [OHTTP Gateway](https://github.com/arminsabouri/ohttp-gateway/tree/secp256k1-KEM-support) locally. Automated tests are still TODO. 

Seeking concept Ack/Nack

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -72,6 +72,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "autocfg"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -281,15 +287,15 @@ checksum = "b65c2e1ab98050c93cc9ada070d5070e014329c65196d03f6f8f1f75c2666f37"
 dependencies = [
  "base64 0.22.1",
  "native-tls",
- "rustls",
- "rustls-native-certs",
- "rustls-webpki",
+ "rustls 0.23.38",
+ "rustls-native-certs 0.8.3",
+ "rustls-webpki 0.103.13",
  "serde",
  "serde_json",
  "tokio",
  "tokio-native-tls",
- "tokio-rustls",
- "webpki-roots",
+ "tokio-rustls 0.26.4",
+ "webpki-roots 1.0.7",
 ]
 
 [[package]]
@@ -327,15 +333,15 @@ checksum = "980479e6fde23246dfb54d47580d66b4e99202e7579c5eaa9fe10ecb5ebd2182"
 
 [[package]]
 name = "byteorder"
-version = "1.4.3"
+version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "14c189c53d098945499cdfa7ecc63567cf3886b3332b312a5b4585d8d3a6a610"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.2.1"
+version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec8a7b6a70fde80372154c65702f00a0f56f3e1c36abbc6c440484be248856db"
+checksum = "d71b6127be86fdcfddb610f7182ac57211d4b18a3e9c82eb2d17662f2227ad6a"
 
 [[package]]
 name = "bzip2"
@@ -561,6 +567,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "data-encoding"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d26bb7bacfa6d2e9d049b98978076e7ab6517caddd4d47b8d53e7f33a85846f8"
+
+[[package]]
 name = "digest"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -610,6 +622,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "equivalent"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88bffebc5d80432c9b140ee17875ff173a8ab62faad5b257da912bd2f6c1c0a1"
+
+[[package]]
 name = "errno"
 version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -630,8 +648,12 @@ dependencies = [
  "electrsd",
  "hex-conservative 0.2.2",
  "hkdf 0.12.2",
+ "http-body-util",
+ "hyper",
+ "hyper-util",
  "log",
  "native-tls",
+ "ohttp-relay",
  "openssl",
  "openssl-sys",
  "security-framework 2.11.1",
@@ -681,6 +703,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "fnv"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6cc484842f1e2884faf56f529f960cc12ad8c71ce96cc7abba0a067c98fee344"
+
+[[package]]
 name = "foreign-types"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


luisschwab:
Concept ACK, this is a cool ideia. We are, however, in the process of completely dropping `reqwest`/`minreq` in favor of [`bitreq`](https://crates.io/crates/bitreq). Can you change the PR to use that?

pseudozach:
I’ve also got a branch that ports this to the current bitreq client and gets the CI passing:
https://github.com/pseudozach/rust-esplora-client/tree/ohttp-bitreq

I also ended up building [a compatible Bitcoin-HPKE OHTTP gateway (obitcoin)](https://github.com/pseudozach/obitcoin) so I could actually do end-to-end testing with the client.

If any of that is useful, I’m happy to fold the gateway-backed smoke test, CI workflow, or related pieces into this PR, or you can cherry-pick whatever is helpful.

arminsabouri:
> I’ve also got a branch that ports this to the current bitreq client and gets the CI passing: https://github.com/pseudozach/rust-esplora-client/tree/ohttp-bitreq

This was useful ref for marking certain deps as duplicates. Thank you.

 >I also ended up building [a compatible Bitcoin-HPKE OHTTP gateway (obitcoin)](https://github.com/pseudozach/obitcoin) so I could actually do end-to-end testing with the client.
> If any of that is useful, I’m happy to fold the gateway-backed smoke test, CI workflow, or related pieces into this PR, or you can cherry-pick whatever is helpful.

I suppose it could reduce the test boilerplate in this PR. i.e removing my dummy gateway. Do you already have a int.  . e2e test that uses your gateway?

pseudozach:
> > I’ve also got a branch that ports this to the current bitreq client and gets the CI passing: https://github.com/pseudozach/rust-esplora-client/tree/ohttp-bitreq
> 
> This was useful ref for marking certain deps as duplicates. Thank you.
> 

happy to hear that!

> 
> I suppose it could reduce the test boilerplate in this PR. i.e removing my dummy gateway. Do you already have a int. . e2e test that uses your gateway?


e2e test is available here: https://github.com/pseudozach/rust-esplora-client/blob/ohttp-bitreq/tests/ohttp_smoke.rs
and runs here: https://github.com/pseudozach/rust-esplora-client/blob/ohttp-bitreq/.github/workflows/ohttp-smoke.yml

arminsabouri:
Cleaned up the commits. Couple things are bothering me.
1. There is alot of boiler plate that could and should be abstracted away. things like bhttp payload construction. Assuming ohttp gets integrated into other clients in this space we could deduplicate this code. I already started putting together a ohttp-client crate. I'll share it later. Probably not a blocker for this PR -- something that could be refactored in a follow up. Maintainers please lmk if you disagree. 
2. `just audit` fails on the second commit bc of the ohttp-relay dep. I could go the route @pseudozach went with (above) however, I would want the test harness to create and clean up the relay/gateway and electrum server instead of CI doing all of that. 

The other thing worth revisiting is depending on ohttp vs bitcoin-ohttp. I dont see a compeling reason to use the secp256k1 curve for hpke. I think the main point that justified bitcoin-ohttp was clients already are using the same crypto primitives (secp256k1, sha256, polychaha). The tradeoff is clients that use the bitcoin-ohttp dep cannot use any of the existing ohttp libraries and/or infra. And this becomes a maintaince headache -- esp as new features comes up such as, PQ support and ohttp-streaming

DanGould:
> the main point that justified bitcoin-ohttp was clients already are using the same crypto primitives

Yes the same primitives AND the same exact dependencies. So no RustCrypto secp256k1, just the same exact deps rust-bitcoin already has everywhere. This does make it incompatible with existing ohttp libraries unless those were PR'd. Even if there was support for other crypto in this client.

Regarding infra, I'm not sure what running OHTTP Infra if any is open for the public to use. My thinking was to use the payjoin infra to bootstrap, along with a new `allowed_purposes` string based on the Directory Interactions spec (https://github.com/bitcoin/bips/blob/master/bip-0077.md#relaydirectory-interactions).

I have not though through post-quantum support or ohttp-streaming story yet because they have not (until now) been demanded by integrators. Lack of demand is a big reason this has been sidelined in the first place. Cake and Bull Bitcoin mobile already sync over Tor by default, so getting OHTTP esplora sync isn't a huge win. It's a bigger win for Cove, though.

arminsabouri:
> Yes the same primitives AND the same exact dependencies. So no RustCrypto secp256k1, just the same exact deps rust-bitcoin already has everywhere. This does make it incompatible with existing ohttp libraries unless those were PR'd. Even if there was support for other crypto in this client.

Why can't AEAD and hash primitives use rust-bitcoin deps? IIUC bitcoin-hpke depends directly on RustCrypto's (or some other crate's) chacha20poly1305, sha2, hkdf, and hmac. 
 
The tradeoff here to consider: is it worth the maintance overhead of maintaining both bitcoin-hpke and bitcoin-ohttp vs clients will need to use new crypto primitives (namely X25119) to use ohttp. And does the latter then exclude some subset of clients from using ohttp for electrum calls (and why? I understand bitcoin-core wallet does not want to depend on X25119 but who else has this constraint?). 

Separately: bitcoin-ohttp's rust-hpke feature currently bundles a legacy RustCrypto-based backend as dead code alongside the bitcoin-hpke backend actually in use. That's the source of the duplicate-crate entries in this diff: https://github.com/bitcoindevkit/rust-esplora-client/pull/145/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R83-R97.

>Regarding infra, I'm not sure what running OHTTP Infra if any is open for the public to use. My thinking was to use the payjoin infra to bootstrap, along with a new allowed_purposes string based on the Directory Interactions spec (https://github.com/bitcoin/bips/blob/master/bip-0077.md#relaydirectory-interactions).

Can the directory proxy requersts to electrum servers today? This may be problematic if the payjoin users are also using the same directory for the electrum gateway.

DanGould:
> Why can't AEAD and hash primitives use rust-bitcoin deps?

They can and should. `chacha20poly1305` crate was a dependency because the rust-bitcoin https://crates.io/crates/chacha20-poly1305 was not yet released. The others were just not demanded until now and it is another hygiene item to do.

> clients will need to use new crypto primitives (namely [X25519]) to use ohttp

Where is this demand coming from? My inclination would be to wait until p2p adopts a post quantum scheme and then use that (edit: assuming you're asking about because you want X25519+MLKEM768? ).

> I understand bitcoin-core wallet does not want to depend on X25119 but who else has this constraint?

Not sure yet. The idea of using the lowest common denominator primitives to remove the entire class of questions / concerns from implementers. 

> bitcoin-ohttp's rust-hpke feature currently bundles a legacy RustCrypto-based backend as dead code

whoops! siccing bots on it overnight. Been a while since either crate has gotten any love

> Can the directory proxy requersts to electrum servers today? 

No, directories don't proxy requests, OHTTP Relays do, and Relays cannot proxy requests to esplora servers until they adopt an [`allowed_purposes` scheme](https://github.com/bitcoindevkit/rust-esplora-client/pull/145/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R83-R97), but the payjoin-mailroom is willing & ready to accept that change and promote it to infrastructure operators in the next update.

edit: Alternatively, an esplora server *could*, (though perhaps shouldn't) be hosted behind an OHTTP Gateway with the payjoin directory BIP 77 specified `allowed_purpose` to opt in to receive requests from any relay willing to forward requests to payjoin directories.

Electrum is off the table because it's direct streaming over TCP socket, not HTTP requests like esplora which can be encapsulated in OHTTP afaiu.

> This may be problematic if the payjoin users are also using the same directory for the electrum gateway.

What problem do you foresee specifically? The relay should see padded-length encrypted payloads (perhaps 1kb) and forward them to esplora servers who have opted in with the `allowed_purposes` advertisement. The directory should not be involved, and if a mailroom serves as both directory and OHTTP Relay, assuming directory requests reach them from OHTTP they won't know the source origin IPs that may also be interested in making esplora requests.

arminsabouri:
> My inclination would be to wait until p2p adopts a post quantum scheme and then use that 

I didnt mean ML-KEM but I also don't agree with this framing. Why should the rest of the bitcoin-ecosystem wait for core to upgrade p2p? 

> (edit: assuming you're asking about because you want X25519+MLKEM768? ). 

no, if clients use the ohttp crate they will use x25119 or p256 for KEM. Depending on the application this may be a new dependency (for bitcoin-core wallet it surely is). The question is, is that enough reason for a client not to use ohttp? The answer seems to be "maybe" until we survey and collect more info. The only client that we know will def not accept new crypto primitives is core -- but they are unlikely to use ohttp for any other reason than bip78. Other clients are probably already using these in the context of other network connections. 

> Been a while since either crate has gotten any love

Another thing: I believe bitcoin-ohttp is has a couple critical cargo audit findings from upstream. Specifically when the server feature is on.

 > No, directories don't proxy requests, OHTTP Relays do, and Relays cannot proxy requests to esplora servers until they adopt an [allowed_purposes scheme](https://github.com/bitcoindevkit/rust-esplora-client/pull/145/changes#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R83-R97), but the payjoin-mailroom is willing & ready to accept that change and promote it to infrastructure operators in the next update.

Esplora servers are not ohttp gateways *yet*. I'm not sure any one is wokring on that. But **I do think its neccecary**. If the ohttp gateway was not the target (Esplora) this allows the gateway operator to cluster addresses by using other metadata / hueristics (timing mainly) but not correlate it to you network id. Unfortunately I don't think we can or should just stand up a gateway infront of esplora server.    

More specifically I am saying the pj directory should not be the ohttp gateway -- because the directory operator alsoi knows the timing of its own target resources.  Re-using the same ohttp relays is probably fine.

PR #153, opened 2026-01-04
fix(ci): bump pinned `webpki-roots` to 1.0.5

oleonardolima:
I was about to merge #114, but there's been a new release (i.e [v1.0.5](https://crates.io/crates/webpki-roots/versions)) for `webpki-roots` which breaks our MSRV CI.

This PR updates its pinned version to fix our MSRV CI jobs, so we can move forward with #114.


luisschwab (approved):
ACK 477cf8adbc31bbf728da1c866142d2293b615950

PR #159, opened 2026-01-07
chore(submit_package): use `unwrap_or_default` instead of `.unwrap()`

oleonardolima:
### Description

- use `unwrap_or_default()` instead of `.unwrap()`, ideally it should return a specific error variant, it'd be a breaking change though and can be added in a follow-up.
- it's only possible to use the internal `minreq::Error` to wrap the `serde_json::Error`, but not the `reqwest::Error` as it does not have public constructors.

### Notes to the reviewers

I personally wanted to remove the `.unwrap()` and properly handle the error, though the options I had in mind:
- (i) adding a new `SerdeJson` variant is a breaking change; we can do it as a follow-up for 0.13.0
- (ii) mapping and wrapping it into either minreq/reqwest errors is only possible in minreq, so I thought it's better to keep it standard and use `.unwrap_or_default()` on both for now.

### DIFF
--- a/src/blocking.rs
+++ b/src/blocking.rs
@@ -144,7 +144,7 @@ impl BlockingClient {
             }
             Ok(resp) => {
                 let hex_str = resp.as_str().map_err(Error::Minreq)?;
-                let hex_vec = Vec::from_hex(hex_str).unwrap();
+                let hex_vec = Vec::from_hex(hex_str)?;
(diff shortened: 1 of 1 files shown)


luisschwab (approved):
ACK c2c5485c67b218d092b35e238a1c54da5692fcda

PR #168, opened 2026-01-31
chore: use the `alloc` feature on `serde_json`

luisschwab:
This PR imports `serde_json` with the `alloc` feature.


oleonardolima:
I did this change [here](https://github.com/oleonardolima/rust-esplora-client/tree/feat/add-serde-json-std-feature), but didn't open the PR because I didn't test if it would actually clash when using std.

oleonardolima (approved):
ACK 5c57ee4c63ece0acac15111df2010707939098fe

PR #186, opened 2026-04-23
chore: upgrade reqwest dependency to 0.13

AmbientTea:
Hello, `reqwest` 0.13 has been out for some time now and `esplora` still using 0.12 conflicts with other crates that have already upgraded. Is there a reason for staying on 0.12? If not, I'm submitting this PR to bump it. 

Changes:
- `async` feature now enables `reqwest/query` - which in 0.13 hides http query support
- `async-https` feature now enables `reqwest/native-tls-no-alpn` instead of `reqwest/default-tls` - in 0.13 `rustls` is the new default and the feature equivalent of the old `default-tls` is now `native-tls-no-alpn`. for the same reason `async-https-rustls` now only pulls `async` and nothing else.
- `reqwest` 0.13 dropped `rustls-tls-manual-roots` feature, same effect for `async-https-rustls-manual-roots` feature is now achieved by calling `builder.tls_certs_only([])` 
- pins `tempfile` to 3.14.0 to preserve msrv


luisschwab:
Thanks, but were migrating to the `bitreq` crate in the near future. No plans on bumping `reqwest`.

AmbientTea:
Good to know, thank you. Any timeline when this change would release?

luisschwab:
We're waiting for a new upstream release on their part, but that should happen soon.

AmbientTea:
Great, thanks!

PR #198, opened 2026-05-01
Use `bitcoin::Amount` and `bitcoin::FeeRate`

### LINKED ISSUE #192: Use `FeeRate` for all fee fields

luisschwab:
Closes #192 

## Changelog
```
## Added
- Add `deserialize_fee_histogram`

## Breaking 
- Use `Amount` and `FeeRate` on `MempoolStats.fee_histogram`
- `submit_package` takes in `FeeRate` for `maxfeerate` instead of `f64`
- `submit_package` takes in `Amount` for `maxburnamount` instead of `f64`
- `get_fee_estimates` returns `FeeRate` instead of `f64`

## Changed
- Update `test_feerate_parsing` with an up to date fee histogram
```

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -286,7 +286,8 @@ pub struct MempoolStats {
     /// An array of `(feerate, vsize)` tuples, where each entry's `vsize` is the total vsize
     /// of [`Transaction`]s paying more than `feerate` but less than the previous entry's `feerate`
     /// (except for the first entry, which has no upper bound).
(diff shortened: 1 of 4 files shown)


oleonardolima (approved):
tACK 291aa096ffa1ec7e1fa9b0659270b9cd4d0885db

I think one point of attention would be that currently the FeeRate in `get_mempool_stats` is being floored (e.g [5.0072994, 50347] -> (FeeRate(1251), Weight(50347))).

PR #203, opened 2026-05-03
Implant `cargo-rbmt`

### LINKED ISSUE #178: Move CI and local testing to `cargo-rbmt`
This will make MSRV management and feature parity between local and CI testing.

I'll improve the README while I'm at it.

luisschwab:
Closes #178 

This PR introduces [`cargo-rbmt`](https://git.rust-bitcoin.org/rust-bitcoin/rust-bitcoin-maintainer-tools/src/branch/master/cargo-rbmt) for local and CI checking, linting, testing and lockfile generation.

## Changelog
```
## Added
- Add vendored `setup-rbmt` action (`.github/actions/setup-rbmt/action.yml`)
- Add new CI workflow using `cargo-rbmt` (`.github/workflows/rust.yml`)
- Add `cargo-rbmt`-specific fields on `Cargo.toml`:
  - Define stable and nightly toolchains
  - Define feature sets for testing
  - Allow duplicate dependencies on the tree (`getrandom`, `windows-sys`, `wit-bindgen`)
- Commit `Cargo-recent.lock` and `Cargo-minimal.lock`
- Add Zizmor CI workflow
- Add dependabot on CI

## Removed
- Remove old `.github/workflows/cont_integration.yml` workflow

## Changed
- Update the `justfile` with `cargo-rbmt` recipes
- Fix feature flags on code
- Pin direct and transitive dependencies for MSRV on `Cargo.toml`
- Fix Zizmor warnings and errors on old CI workflows
- Update the `README.md`

### CODE REVIEW

File: .clippy.toml [hunk 1]
```diff
@@ -1 +0,0 @@
-msrv="1.75.0"
```
oleonardolima, line 1: nit: i think we'd need to keep this one, for compatibility with users not using `cargo-rbmt`.

Cargo.toml
```diff
@@ -10,41 +10,78 @@ license = "MIT OR Apache-2.0"
...
-serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0", default-features = false }
+serde = { version = "1.0.194", features = ["derive"] }
+serde_json = { version = "1.0.117", features = ["std"], default-features = false }
 bitcoin = { version = "0.32", features = ["serde", "std"], default-features = false }
-hex = { version = "0.2", package = "hex-conservative" }
-log = "^0.4"
-bitreq = { version = "0.3.4", optional = true }
+hex = { version = "0.2.2", package = "hex-conservative" }
+bitreq = { version = "0.3.5", optional = true }
```
oleonardolima, line 27: nit: now having another look at it it would be better to have both `proxy` and `json-using-serde` setup here, they're always being used.

justfile
```diff
@@ -1,47 +1,75 @@
+alias a := audit
 alias b := build
 alias c := check
+alias cs := check-sigs
```
oleonardolima, line 4: nit: I'd suggest `verify-sigs`, it could be confused with `cargo check` one.

justfile
```diff
@@ -1,47 +1,75 @@
...
+
+[doc: "Run pre-push checks"]
+pre-push:
+    @just lock
+    @just check
+    @just doc
+    @just test
+    @just audit
+    @just zizmor
+    @just check-sigs
```
oleonardolima, line 75: I'd avoid having `zizmor` in here, as it'd break for user that does not have `uvx` installed and doesn't want to install it.

Another alternative is only running zizmor above when uvx is available, warning but not breaking otherwise.

src/lib.rs
```diff
@@ -326,11 +337,11 @@ impl_error!(bitcoin::consensus::encode::Error, BitcoinEncoding, Error);
 impl_error!(bitcoin::hex::HexToArrayError, HexToArray, Error);
 impl_error!(bitcoin::hex::HexToBytesError, HexToBytes, Error);
 
+#[cfg(all(test, feature = "blocking", feature = "async", feature = "tokio"))]
```
oleonardolima, line 340: nice!

oleonardolima (approved):
utACK 8fc927361d9f88ad16dac83f8f128361547776bd

I left a few comments that can be addressed in a follow-up.

oleonardolima:
ping @luisschwab :shipit:

.clippy.toml
luisschwab: We can't keep this since it conflicts with `cargo-rbmt`.

Cargo.toml
luisschwab: We can look into this in a follow up.

justfile
luisschwab: It's simple enough to install `uv` via `cargo`: `~$ cargo install uv`.

Issue #210, opened 2026-05-30
Support mempool.space API changes


notmandatory:
The mempool.space APIs are diverging from the original ones from blockstream. We need to figure out a way to support both.

### API Docs

mempool.space: https://mempool.space/docs/api/rest
blockstream:  https://github.com/Blockstream/esplora/blob/master/API.md

notmandatory:
This issue came up today because @MaxFangX had a production issue because mempool.space removed the `/fee-estimates` API.  Mempool may re-enable the API to give end-users time to update their code, but still TBD on that.

See: https://discord.com/channels/753336465005608961/1359516227235741969/1510389629180969091

orangesurf:
/fee-estimates will continue to be supported but we recommend using the precise fees API endpoint https://mempool.space/api/v1/fees/precise

luisschwab:
It would be great if you and Blockstream would come up with a versioning scheme and endpoint like the Electrum Protocol does it.

oleonardolima:
I feel that having the new trait I discussed with @luisschwab (#200) would benefit this, we can have strict ones, like: `EsploraMethods` and `MempoolSpaceMethods`.

luisschwab:
> I feel that having the new trait I discussed with @luisschwab (https://github.com/bitcoindevkit/rust-esplora-client/issues/200) would benefit this, we can have strict ones, like: EsploraMethods and MempoolSpaceMethods.

Indeed

tnull:
> It would be great if you and Blockstream would come up with a versioning scheme and endpoint like the Electrum Protocol does it.

Fully agree with this sentiment. Given that Blockstream's version is likely not seeing many changes, it could be consider `v0` and any mempool-specific changes could be easily added in a backwards compatible manner as part of a new `v1` that is simply communicated via an HTTP header field?

luisschwab:
@orangesurf you're still breaking the API:

```shell
~$curl -s https://mempool.space/api/fee-estimates | jq
{
  "1": 1.975,
  "2": 1.975,
  "3": 1.28,
  "4": 1.28,
  "5": 1.28,
  "6": 0.743,
  "7": 0.743,
  "8": 0.743,
  "9": 0.743,
  "10": 0.743,
  "11": 0.743,
  "12": 0.743,
  "13": 0.743,
  "14": 0.743,
  "15": 0.743,
  "16": 0.743,
  "17": 0.743,
  "18": 0.743,
  "19": 0.743,
[...]

Issue #200, opened 2026-05-02
Implement a super-trait for client methods


luisschwab:

luisschwab:
ping @oleonardolima :shipit:

luisschwab:
We could try to tackle this when making the clients sans-IO.

PR #212, opened 2026-06-04
fix: tolerate non-numeric keys in get_fee_estimates (mempool.space deprecation warning)

Copexit:
mempool.space added a non-numeric `"warning"` key to its (deprecated) `/fee-estimates` response:

```json
{"1":4.8, ... ,"1008":0.1,"warning":"This endpoint is deprecated ... use /api/v1/fees/precise"}
```

`get_fee_estimates` deserializes the body as `HashMap<u16, f64>`, so the `"warning"` key fails to parse and the whole call errors. Any client pointed at a mempool.space Esplora is affected.

The most severe downstream impact is **ldk-node**: it calls `get_fee_estimates` during `Node::start()`, so the failed parse makes the node refuse to start and crash-loop on restart. This is a real production failure mode (see #210).

### Fix

Parse `/fee-estimates` leniently: deserialize into a permissive map and keep only entries whose key is a numeric confirmation target and whose value is a number; ignore anything else (such as the `"warning"` string). Behavior is unchanged for spec-compliant servers (e.g. Blockstream). Applied to both the async and blocking clients.

### Changelog

- Fixed: `get_fee_estimates` tolerates non-numeric entries in the `/fee-estimates` response (e.g. mempool.space deprecation `warning`), which previously broke fee estimation for mempool.space-backed clients and prevented ldk-node from starting.

Refs #210

### DIFF
--- a/src/async.rs
+++ b/src/async.rs
@@ -545,7 +545,18 @@ impl<S: Sleeper> AsyncClient<S> {
     /// Returns a [`HashMap`] where the key is the confirmation target in blocks
     /// and the value is the estimated [`FeeRate`].
     pub async fn get_fee_estimates(&self) -> Result<HashMap<u16, FeeRate>, Error> {
-        let estimates_raw: HashMap<u16, f64> = self.get_response_json("/fee-estimates").await?;
+        // Tolerate non-numeric entries in the response. mempool.space added a
+        // non-numeric `"warning"` key to its (deprecated) `/fee-estimates`
+        // endpoint, which makes a strict `HashMap<u16, f64>` deserialization
+        // fail and breaks every downstream caller (e.g. ldk-node refuses to
+        // start). Keep only entries whose key is a numeric confirmation target
+        // and whose value is a number.
+        let raw: HashMap<String, serde_json::Value> =
+            self.get_response_json("/fee-estimates").await?;
+        let estimates_raw: HashMap<u16, f64> = raw
+            .into_iter()
+            .filter_map(|(target, feerate)| Some((target.parse::<u16>().ok()?, feerate.as_f64()?)))
+            .collect();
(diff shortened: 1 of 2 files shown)


luisschwab:
NACK. This client is not responsible for conforming to server implementations that deviate from the specification (https://github.com/Blockstream/esplora/blob/master/API.md#get-fee-estimates). 

Instead, Mempool should return `301 Moved Permanently`/`303 See Other`/`308 Permanent Redirect`.

Copexit:
Thanks for the review, and fair point on spec purity.

The practical wrinkle: per #210, the mempool team has said they're **keeping** `/fee-estimates` (recommending `/api/v1/fees/precise` but not removing the old one), and they added the `warning` key rather than redirecting, so a 301/308 from mempool doesn't look like it's coming. In the meantime this breaks `node.start()` in ldk-node for anyone pointed at a mempool.space backend, which is a very common configuration.

This change is strictly more lenient, Blockstream/spec-compliant responses parse identically; it only drops non-numeric keys. So there's no spec regression for compliant servers. But I'm happy to defer to whatever direction you prefer: fold it into the `EsploraMethods` / `MempoolSpaceMethods` split discussed in #200, gate it behind a `mempool` feature, or close this if you'd rather solve it there. Mostly flagging that real users of downstream crates (ldk-node) are hitting node-startup failures today.

junderw:
https://github.com/bitcoindevkit/rust-esplora-client/issues/210#issuecomment-4637292970

As mentioned in this comment I made. We should be pushing a fix shortly.

However, most wallets will be converting "1" to something human readable like "fast" anyways, so it would be best if end users add a normal HTTP client to their apps (use the same version that bdk uses if you don't want to add a new dependency) and just pull Mempool's precise fee API values.

Copexit:
Closing as https://github.com/mempool/mempool/pull/6546 has been merged.

PR #242, opened 2026-08-16
ci: audit all lockfiles and bump cargo-rbmt to v0.5.3

### LINKED ISSUE #227: ci: bump `cargo-rbmt` to v0.5.2

### LINKED ISSUE #228: ci: audit all lockfiles

EliteCoder18:
## Description

Closes #228  
Closes #227

Updates the nightly audit workflow to check all three dependency resolutions used by the repository:

- A freshly updated `Cargo.lock`
- The committed `Cargo-recent.lock`
- The committed `Cargo-minimal.lock`

The audit jobs run sequentially to avoid multiple matrix jobs trying to create or update the same RustSec advisory issue simultaneously.

Also bumps `cargo-rbmt` to v0.5.3, updates the affected rbmt command syntax, installs the tools declared in `Cargo.toml`, pins `cargo-audit` to v0.22.2, and bumps Zizmor to v1.29.0.

The local `just audit` recipe now mirrors CI by auditing all three lockfiles.

## Notes to the reviewers

- `Cargo.lock` remains uncommitted and is freshly updated before auditing. The recent and minimal resolutions continue to use the committed rbmt lockfiles.

- The audit matrix uses `max-parallel: 1` so `rustsec/audit-check` jobs do not race while reporting the same advisory through GitHub Checks or issues.

- `cargo audit` flagged vulnerable versions of `bzip2`, `mio`, and `ring` in `Cargo-minimal.lock`, so minimum safe development dependency constraints were added to prevent those versions from being selected.

- The rbmt v0.5.3 migration includes:
  - `cargo rbmt docsrs` → `cargo rbmt docs`
  - `--lock-file` → `--lockfile`
  - Installing declared tools with `cargo rbmt tools`
 
- `electrsd` was bumped to v0.41.0 in a separate commit. This did not remove the need for the `bzip2 >= 0.4.4` constraint because minimal resolution still selects vulnerable `bzip2 0.4.3` without it.

## Changelog
```
* Audit fresh, recent, and minimal lockfiles
* Bump `cargo-rbmt` to v0.5.3
* Bump `electrsd` to v0.41.0
* Patch all RUSTSEC warnings and vulnerabilities
* Bump stable and nightly toolchains
```


.github/actions/setup-rbmt/action.yml
```diff
@@ -12,3 +12,7 @@ runs:
         - name: Install Rust Toolchains
           shell: bash
           run: cargo rbmt toolchains
+
+        - name: Install cargo-rbmt Tools
+          shell: bash
+          run: cargo rbmt tools
```
luisschwab, line 21: Can you get rid of the `rbmt-version` file and declare the version directly on `Cargo.toml`?

ref: https://github.com/luisschwab/bdk-floresta/blob/master/.github/actions/setup-rbmt/action.yml
ref: https://github.com/luisschwab/bdk-floresta/blob/master/Cargo.toml#L56-L57

.github/workflows/rust.yml
```diff
@@ -22,11 +22,11 @@ jobs:
                     - name: Lint Code
                       task: lint
                     - name: Build Documentation
-                      task: docsrs
+                      task: docs
```
luisschwab: Unrelated nit:

```suggestion
                    - name: Documentation
                      task: docs
                    - name: Formatting
                      task: fmt --check
                    - name: Linting
                      task: lint
```

Cargo.toml
```diff
@@ -60,15 +63,17 @@ blocking-https-rustls-probe = ["blocking", "bitreq/https-rustls-probe"]
 stable = "1.96.1"
 nightly = "nightly-2026-07-03"
 
+[package.metadata.rbmt.tools]
+cargo-audit = "=0.22.2"
```
luisschwab: The `=` makes no difference in this case

```suggestion
cargo-audit = "0.22.2"
```

luisschwab (changes requested):
You can also bump `electrsd` to v0.41.0 on this PR, maybe that will remove the need for the `bzip2` pin

Cargo.toml
luisschwab: While you're at it, you could put `zizmor` under `rbmt.tools` and update the zizmor job as well.

EliteCoder18:
Addressed the review feedback in the latest commit:
- moved the `RBMT` version and managed tool versions into `Cargo.toml`
- updated the `Zizmor` workflow to use RBMT-managed zizmor
- bumped `electrsd` to `0.41.0`
- regenerated the `lockfiles` and retained the minimal-resolution safety pins required for `cargo audit`
I also checked whether the `electrsd` bump removes the `bzip2` constraint. It does not: minimal resolution still selects vulnerable `bzip2 0.4.3` without the constraint, so the `bzip2 >= 0.4.4` pin remains necessary for `cargo audit`.

luisschwab:
Can you split the `electrsd` bump on a separate commit?

luisschwab:
There's a conflict, and it should be 3 commits, each one doing a single thing.

luisschwab:
@EliteCoder18 there should be 3 commits, each doing one thing. And there's a conflict that needs resolving.

luisschwab:
Seems like `cargo-rbmt` v0.5.3 was just released. Can you bump it to that instead?

luisschwab:
Can you update `zizmor` to the latest? Just run `cargo rbmt tools --update`

luisschwab (approved):
ACK 8832f36883bf96c593741051e677be3840e8ec23

luisschwab:
Merging this one once GitHub get's the API back online

luisschwab:
Merging this one once GitHub gets the API back online

luisschwab (approved):
self-ACK de476a77eab0c8d7e6f92a4d59dd85cd416013c5

Issue #241, opened 2026-08-14
Allow opt-in `reqwest` through a `async-reqwest` feature


MaxFangX:
Pursuant to https://github.com/rust-bitcoin/corepc/issues/587, would BDK be open to adding an opt-in `async-reqwest` feature which allows users (like us) to opt-in to using `reqwest` instead of `bitreq`? `bitreq` appears unlikely to implement HTTP/2, so allowing `reqwest` seems to be the only way to let users use HTTP/2. We have it implemented already (it's about ~260 LOC) -  just wondering whether you'd be OK with this being upstreamed.

`bitreq` would still be the default client, but users can set `default-features = false` to disable `bitreq` and set `async-reqwest` to use `reqwest` instead.

luisschwab:
I don't see this landing here. However, there are future plans to make the clients sans-io, so you could implement whatever transport you want.

MaxFangX:
Yes, some sort of generic interface for the transport would be perfect!

MaxFangX:
> However, there are future plans to make the clients sans-io, so you could implement whatever transport you want.

Looks like https://github.com/bitcoindevkit/rust-esplora-client/issues/97 is the place where that is documented. Closing this issue accordingly.

Issue #97, opened 2024-09-06
Refactor library so that it can work with any HTTP client crate


evanlinjin:
## The Problem

Right now `rust-esplora-client` is designed to only work with HTTP client crates that are baked in (i.e. `minreq` for blocking, `reqwest` for async). However, these HTTP client crates may not be preferred by the caller. The caller may also wish to use a different version of the HTTP client crate.

## Proposed Solution

Represent each endpoint call as a distinct type. Introduce a trait that is to be implemented for each endpoint call type. This trait will contain the *template* of how to form the HTTP request and parse the HTTP response body and status code.

```rust
pub trait HttpRequest {
    type Output: for<'a> serde::de::Deserialize<'a>;

    fn request_url_path(&self) -> String;
    fn request_method(&self) -> &str;
    fn request_body(&self) -> Vec<u8>;
    fn parse_response<E>(status: i32, body: &[u8]) -> Result<Self::Output, ResponseError<E>>;
}

pub enum ResponseError<E> {
    Json(serde_json::Error),
    HttpStatus { status: i32, body: Vec<u8> },
    Client(E),
}
```

Examples of endpoint types...

```rust
pub struct GetTx {
    pub txid: Txid,
}

impl HttpRequest for GetTx {
    type Output = Option<Transaction>;

    fn request_url_path(&self) -> String {
        format!("/tx/{}/raw", self.txid)
    }

    fn request_method(&self) -> &str {
        "GET"
    }

    fn request_body(&self) -> Vec<u8> {
        Vec::with_capacity(0)
    }

    fn parse_response<E>(status: i32, body: &[u8]) -> Result<Self::Output, ResponseError<E>> {
        match status {
            200 => Ok(serde_json::from_slice(body).map_err(ResponseError::Json)?),
            404 => Ok(None),
            error_status => Err(ResponseError::HttpStatus {
                status: error_status,
                body: body.to_vec(),
            }),
        }
    }
}

pub struct PostTx {
    pub tx: Transaction,
}

impl HttpRequest for PostTx {
    type Output = ();

    fn request_url_path(&self) -> String {
        "/tx".to_string()
    }

    fn request_method(&self) -> &str {
        "POST"
    }

    fn request_body(&self) -> Vec<u8> {
        bitcoin::consensus::encode::serialize(&self.tx)
            .to_lower_hex_string()
            .as_bytes()
            .to_vec()
    }

[...]

PR #243, opened 2026-08-18
refactor!: normalize responses into a shared HttpResponse

MaxFangX:
#97 plans to abstract the HTTP client away. This PR makes progress towards
that goal by centralizing response handling in one client-agnostic type, so
that only the code producing it touches the underlying HTTP client.

Additional motivation: To address the immediate performance and reliability
problems caused by the switch to `bitreq` (outlined in rust-bitcoin/corepc#587),
Lexe needs to switch back to using `reqwest` as our underlying HTTP client.
After #241, it looks like a patch to add `reqwest` back in under an opt-in
feature flag unlikely to be upstreamed, so Lexe will need to maintain a fork of
`rust-esplora-client`. This refactor has minimal logic changes but makes the
maintenance of our fork a bit easier for us.

Concretely, responses are normalized into a private `HttpResponse` (status +
fully-read body) as soon as they leave the HTTP client. Retry loops, status
checking, and all endpoint methods in both clients now operate on this shared
type, so swapping the HTTP client (whether for our reqwest patch or for #97)
only touches the few functions that produce an `HttpResponse`.

The refactor also DRYs up the status checking and error construction previously
duplicated across both async and blocking clients, and deletes the
`bitreq::Response` status predicates, four of which were dead code.
Net -34 lines.

Breaking changes:

- `BlockingClient::post_request` is now private (it returned a raw
  `bitreq::Response`).
- JSON and UTF-8 decode errors are now reported via `Error::SerdeJson` and a
  new `Error::InvalidUtf8` variant rather than wrapped bitreq errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Changed
 
+* refactor(client)!: normalize responses into a shared `HttpResponse` [#243]
 * feat(client): enable request pipelining for `AsyncClient` [#245]
 * chore!: remove deprecated `BlockSummary` and `get_block` [#225]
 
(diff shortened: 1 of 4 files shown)


luisschwab:
This makes sense, but will need to wait until v0.14.0 since it's a breaking change.

